PR #20, opened 2019-10-12
Terminal/iTerm 2 - Fix a resize bug regarding not resizing to a 2/3 & 1/3 (top & bottom) 

Easen:
…cking if a rect is centered within another rect

Fixes the issue of iTerm/Terminal windows not being resized down to a 1/3 & 2/3 due their own resize logic which rounds up/down to the height of a line

### DIFF
--- a/Rectangle/WindowCalculation/BottomHalfCalculation.swift
+++ b/Rectangle/WindowCalculation/BottomHalfCalculation.swift
@@ -10,28 +10,31 @@ import Foundation
 
 class BottomHalfCalculation: WindowCalculation {
     
-    func calculateRect(_ windowRect: CGRect, visibleFrameOfScreen: CGRect, action: WindowAction) -> CGRect? {
+    func calculateRect(_ windowRect: CGRect, lastAction: RectangleAction?, visibleFrameOfScreen: CGRect, action: WindowAction) -> CGRect? {
         
         var oneHalfRect = visibleFrameOfScreen
         oneHalfRect.size.height = floor(oneHalfRect.height / 2.0)
         
-        if Defaults.subsequentExecutionMode.value != .none {
-            if abs(windowRect.midX - oneHalfRect.midX) <= 1.0 {
-                var twoThirdsRect = oneHalfRect
-                twoThirdsRect.size.height = floor(visibleFrameOfScreen.height * 2 / 3.0)
-                
-                if rectCenteredWithinRect(oneHalfRect, windowRect) {
-                    return twoThirdsRect
-                }
-                
-                if rectCenteredWithinRect(twoThirdsRect, windowRect) {
-                    var oneThirdRect = oneHalfRect
-                    oneThirdRect.size.height = floor(visibleFrameOfScreen.height / 3.0)
-                    return oneThirdRect
-                }
-            }
+        if Defaults.subsequentExecutionMode.value == .none {
+            return oneHalfRect
         }
         
-        return oneHalfRect
+        let count = (lastAction?.action == action) ? (lastAction?.count ?? 0) : 0
+        let position = count % 3
+        
+        switch (position) {
+            case 2:
+                var oneThirdRect = oneHalfRect
+                oneThirdRect.size.height = floor(visibleFrameOfScreen.height / 3.0)
+                return oneThirdRect
+            
+            case 1:
+                var twoThirdsRect = oneHalfRect
+                twoThirdsRect.size.height = floor(visibleFrameOfScreen.height * 2 / 3.0)
+                return twoThirdsRect
+            
+            default:
+                return oneHalfRect
+        }
     }
 }
--- a/Rectangle/WindowCalculation/TopHalfCalculation.swift
+++ b/Rectangle/WindowCalculation/TopHalfCalculation.swift
@@ -10,31 +10,35 @@ import Foundation
 
 class TopHalfCalculation: WindowCalculation {
     
-    func calculateRect(_ windowRect: CGRect, visibleFrameOfScreen: CGRect, action: WindowAction) -> CGRect? {
+    func calculateRect(_ windowRect: CGRect, lastAction: RectangleAction?, visibleFrameOfScreen: CGRect, action: WindowAction) -> CGRect? {
         
         var oneHalfRect = visibleFrameOfScreen
         oneHalfRect.size.height = floor(oneHalfRect.height / 2.0)
         oneHalfRect.origin.y += oneHalfRect.height + (visibleFrameOfScreen.height.truncatingRemainder(dividingBy: 2.0))
 
-        if Defaults.subsequentExecutionMode.value != .none {
-            if abs(windowRect.midX - oneHalfRect.midX) <= 1.0 {
+        if Defaults.subsequentExecutionMode.value == .none {
+            return oneHalfRect
+        }
+        
+        let count = (lastAction?.action == action) ? (lastAction?.count ?? 0) : 0
+        let position = count % 3
+        
+        switch (position) {
+            case 2:
+                var oneThirdRect = oneHalfRect
+                oneThirdRect.size.height = floor(visibleFrameOfScreen.height / 3.0)
+                oneThirdRect.origin.y = visibleFrameOfScreen.origin.y + visibleFrameOfScreen.height - oneThirdRect.height
+                return oneThirdRect
+            
+            case 1:
                 var twoThirdsRect = oneHalfRect
                 twoThirdsRect.size.height = floor(visibleFrameOfScreen.height * 2 / 3.0)
                 twoThirdsRect.origin.y = visibleFrameOfScreen.origin.y + visibleFrameOfScreen.height - twoThirdsRect.height
-                if rectCenteredWithinRect(oneHalfRect, windowRect) {
-                    return twoThirdsRect
-                }
-                if rectCenteredWithinRect(twoThirdsRect, windowRect) {
-                    var oneThirdRect = oneHalfRect
-                    oneThirdRect.size.height = floor(visibleFrameOfScreen.height / 3.0)
-                    oneThirdRect.origin.y = visibleFrameOfScreen.origin.y + visibleFrameOfScreen.height - oneThirdRect.height
-                    return oneThirdRect
-                }
-            }
+                return twoThirdsRect
+            
+            default:
+                return oneHalfRect
         }
-        
-        return oneHalfRect
-        
     }
     
 }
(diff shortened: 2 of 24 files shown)


rxhanson:
Thanks for looking into it and making this pull request! I think ultimately we'll want to reference the last window action that's passed to each calculation and use that to determine what the behavior should be, in addition to detecting the current size/pos of the window. That way we're not guessing what a window's tolerance would be. I'll be AFK for a few days, but I might merge this in as a quick fix before taking a look at passing the last window action details around. That change will be a bit more involved, since the last window action doesn't currently have a property for the 2/3 or 1/3 behavior.

Even if I don't merge this in, it counts as a code contribution - shoot me a direct message on gitter.im from your github profile if you wanted a license for Multitouch.

Easen:
Thanks for the feedback (and thanks for the Multitouch license offer too).

I thought that might be an option, tracking the last resize action against the window.  Now looking a bit deeper see the `WindowHistory` class. I

Swift is not something I've really played about with, hence the 'quick fix'.  I might see if I can use that instead of increasing the error margin, just as a learning exercise if thats ok?

rxhanson:
👍Feel free to give it a go. It’ll be a while before I can dig into it, so if you can figure it out that would be great!

Easen:
I've now introduced a "context" which is passed into and returned from each calculation. This is then used to work out what the next position should be by using the previous value.

rxhanson:
Thanks for working on this. I'll review this in a bit!

Rectangle/WindowManager.swift
```diff
@@ -98,13 +99,16 @@ class WindowManager {
         }
 
         let resultingRect = frontmostWindowElement.rectOfElement()
-        windowHistory.lastRectangleActions[windowId] = RectangleAction(action: calcResult.resultingAction, rect: resultingRect)
+        windowHistory.lastRectangleActions[windowId] = RectangleAction(action: calcResult.resultingAction, rect: resultingRect, context: calcResult.context)
     }
 }
 
+typealias ActionContext = [String: Any]
```
rxhanson: Let's see, where to start on this. Swift is a strongly typed language with some pretty useful object oriented concepts built in. By creating a type aliased dictionary with values any type, you're basically building an object that doesn't make use of any of Swift's strongly typed or object oriented (and potentially protocol-based) concepts. This typically makes for code that is unnecessarily difficult to read and maintain. By using the `Any` keyword, any time we want to obtain a value out of that dictionary, we will have to type check/cast it into what we expect it to be. As you could imagine, `Any` is something I always avoid because I almost always want typed objects if I can help it. Instead of a type aliased dictionary, I think what we really want is either a struct or a class that we can pass around. Lucky for us, we already have the `RectangleAction` below. 

Instead of a context object, let's add what we really want to keep track of straight to that struct, since it's already being passed into a `windowCalculation`. What do we want to keep track of? Looks like you built the SupportedPositions enum. Going with an enum for this approach is a good choice, and it's how I would have done it as well. However, let's take a step back and see if there's something better we can keep state of. Perhaps we can keep state without having to return anything extra from the window calculation. This would simplify our code a good bit, as there's not as many pieces scattered around that are giving us this state stored in the last window action. 

My thought here is that instead of keeping track of the SupportedPosition that is determined by the window calculation, we would rather keep track of a "Repeated Execution Count" of the window action. Inside of the WindowManager is where most all of our important logic is stored, and this is something that we can viably keep track of here. If we track this count as part of our `RectangleAction` and pass it along to the `calculateRect` in each `WindowCalculation`, then we can just as easily tell which calculation to do - 1/3, 2/3, etc. Also, we're not binding WindowActions to the enumeration. This is important because I plan on implementing next/prev thirds as part of the first third or last third actions. By passing each WindowCalculation a count instead of an enum, each action doesn't need to be aware of all the other cases in that enum - they have the count and they can `switch` on it to choose what to do.

I'll stop my review here for now - since changes from this comment would affect all the other changes that you have currently in the PR. If you want to keep working on this and have the time, by all means, go for it and I'll keep reviewing. If not - that's ok, too! I don't mind rolling it out.

Easen: Thanks for your feedback/input, really appreciate you taking the time to review and respond.

The reason for my choice of a context with `Any` stems back from C/C++ days I guess, where a `void**` was passed into your method allowing your method store _it’s_ state in how _it_ saw fit, as that method was only going to be only thing that will be accessing it. Which absolutely agree, isn’t very type-safe.

I do like the idea of just keeping an execution count that is passed around. It's a lot cleaner and simpler, I'm annoyed that I didn't see, I think perhaps I was focus on the concept of a ‘context’/‘state’ driven solution. (I think I just got carried away with some of Swift's other cool features, multiple return values, etc.) 

Looking at it with fresh eyes I really don't want this to be merged in in its current state. It needs to be refactored…

Rectangle/WindowCalculation/BottomHalfCalculation.swift
```diff
@@ -8,30 +8,37 @@
 
 import Foundation
 
+private enum SupportedPositions: Int {
```
rxhanson: Typically, I like using a lot of enums to clarify things and get rid of magic numbers. But in this case and any others like it, I like just straight up using the count . It's not hard to figure out what the count means without the enum, especially for anyone that uses the app.

Rectangle/WindowCalculation/BottomHalfCalculation.swift
```diff
@@ -8,30 +8,37 @@
...
         var oneHalfRect = visibleFrameOfScreen
         oneHalfRect.size.height = floor(oneHalfRect.height / 2.0)
         
-        if Defaults.subsequentExecutionMode.value != .none {
-            if abs(windowRect.midX - oneHalfRect.midX) <= 1.0 {
+        if Defaults.subsequentExecutionMode.value == .none {
+            return oneHalfRect
+        }
+        
+        let count = (lastAction?.action == action) ? lastAction!.count : 0
```
rxhanson: Perhaps I should write some coding guidelines for the project (you're the first one to encounter my nit picking). 

I never force unwrap. There is pretty much never a reason why I would want my app to crash if an optional is null. Even though that's not possible in this case, there may come a day when calculateRect gets refactored to take an optional WindowAction, and then we're wondering why it crashed somewhere down the line. As a user, I prefer having something just unexpectedly not happen in the app instead of having it crash seemingly randomly.

I think I would prefer this to look like this:
```
let count: Int? = lastAction?.action == action
    ? lastAction?.count
    : 0
```

Of course, then we can't make use of a switch statement below, but that's not a huge deal to me since we don't have a ton of cases.

Rectangle/WindowCalculation/TopHalfCalculation.swift
```diff
@@ -8,33 +8,41 @@
 
 import Foundation
 
+private enum SupportedPositions: Int {
```
rxhanson: Identical comments to the BottomHalfCalculation.swift

Rectangle/WindowManager.swift
```diff
@@ -98,13 +98,19 @@ class WindowManager {
         }
 
         let resultingRect = frontmostWindowElement.rectOfElement()
-        windowHistory.lastRectangleActions[windowId] = RectangleAction(action: calcResult.resultingAction, rect: resultingRect)
+        let count = (lastRectangleAction?.action == calcResult.resultingAction) ? lastRectangleAction!.count + 1 : 1
```
rxhanson: Again, I would prefer not to force unwrap, even though this is another scenario where lastRectangleAction would not be nil. 
Maybe something like this 

```
var count = 1
if let lastRectangleAction = lastRectangleAction {
    if lastRectangleAction.action = calcResult.resultingAction {
        count = lastRectangleAction.count + 1
    }
}
```

It's kind of annoying to have the `if let` assignment to sift through a potential nil, but it's not hard to get used to seeing it, and in my opinion it just beats force unwrapping. The reason why I'm so cautious about force unwrapping is that it's bit me several times before. When you're dealing with certain APIs, like a lot of C APIs (which we're not here), sometimes values can become nil without you knowing it. I had a lot of weird crashes in Multitouch and all of it came down to force unwrapping. Basically, my thought is just err on the side of no crashes.

Rectangle/WindowCalculation/BottomHalfCalculation.swift
Easen: I like them, as I hate magic numbers, but I see your point the code is simple enough to read without them.

Rectangle/WindowCalculation/BottomHalfCalculation.swift
Easen: I originally had a a `if let` statement to unwrap etc., but then I rewrote it to one line as I didn't see how it was possible to throw an error from there as the isn't optional on the struct. However I see from reading online the swift compiler isn't that smart compared to others and if you use '!' the compiler won't double check and give you a warning.

Instead of using `var` what do you think of using the nil-coalescing operator to safely unwrap the optional?

`let count = (lastAction?.action == action) ? (lastAction?.count ?? 0) : 0`

Rectangle/WindowManager.swift
Easen: I look at your sample code and I just think it's overly verbose for what you're trying to do, 5 lines to safely unwrap a `Optional<>`, I really wish the language had the `.orElse()` Java Optional method (https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElse-T-)

Again what about:

`let count = (lastRectangleAction?.action == calcResult.resultingAction) ? (lastRectangleAction?.count ?? 0) + 1 : 1`

Easen: I've realised you can extend the languages Optional, meaning a `getOrElse()` method (similar to Java .orElse() method) is possible! 

Might reduce the boiler-plate Swift Optional unwrapping

```
extension Optional  {
    func getOrElse<T>(_ value: T) -> T {
        if self != nil {
            return self as! T
        }
        return value
    }
}
```

Rectangle/WindowCalculation/BottomHalfCalculation.swift
rxhanson: Sure, that works for me.

Rectangle/WindowManager.swift
rxhanson: I'm good with your suggestion on providing default values with `??`.  

Although it's nice to be able to extend language functionality, I typically avoid it because it's one of those things that can get out of hand if overused. Switching to different projects and lugging around your suitcase of language extensions isn't ideal.

PR #34, opened 2019-11-21
Pull request for Italian l18n

e-magon:
I've improved the Italian translation, let me know if there are some problems.

Thank you


e-magon:
Ok I will, thank you

PR #42, opened 2019-11-29
Update Main.strings

rafo:
Fixed some language mistakes and logical problems around "Frist two thirds" vs. "Erste zwei drittel" which would no one say like that. Since the thirds are managed around the "middle" aka "Mitte" it makes more sense to translate them "Erstes" und "Letztes" to "Links" und "rechts". Especially since "Links" und "Rechts" are used thouout Rectangle.

Fixed some translations around "Systempreferences" to actual tranlations in MacOS 10.15.1.

Right now i dont understand the meaning of "half actions": "This applies to move or half actions only. Half actions will not  perform 2/3 & 1/3 behavior when enabled." The german translation seems to be wrong...

Cheers!


rafo:
Half to thirds: Ah ok! So it means "Cycle (or "toggle" (but maybe toggle is only between two states)) on the left from 1/2 to 2/3 to 1/3" or, if Traverse is checked, "Cycle through (or "to") Monitor(s) on the left"

rxhanson:
Exactly! I think I'll change the English version to use "cycle" as well.

rafo:
That was my goal ;D I think i could provide a better German translation for this. I might be not as short, but, to be honest, thats often the case with German tranlations...

rafo:
This is the translation:

**Monitorwechsel bei Befehlswiederholung**
Wechsle bei Wiederholung von links oder rechts Befehlen zum nächsten Monitor statt zwischen den Größen ½, ⅔, und ⅓.

rxhanson:
Ok, I've updated the English text for "Cycling displays". I've also merged in your suggested translations for it on your branch.

On First and Last Thirds wording, I would still prefer to leave First & Last instead of Left & Right. If we were to do this, what would the best wording be for the thirds actions?

rxhanson:
I'm going to go ahead and merge this, as I'd like to go ahead and push out another release and I think it's better to have your translations all rolled in than to wait until changes are made for first/last as opposed to left/right. If you don't mind revisiting that, it would be much appreciated! As per my contribution policy, you're welcome to a free Multitouch license for your help - just send me a direct message on https://gitter.im.

rafo:
> Ok, I've updated the English text for "Cycling displays". I've also merged in your suggested translations for it on your branch.
> 
> On First and Last Thirds wording, I would still prefer to leave First & Last instead of Left & Right. If we were to do this, what would the best wording be for the thirds actions?

One solution could be:

Erstes Drittel
Zwei Drittel zuerst
Mittleres Drittel
Zwei Drittel zuletzt
Letztes Drittel

It still sounds "wrong" or "not so intuitive" intuitive in german - meaning that you are completely lost without looking at the icons.Which led me to think that, with the help of the icons, either way could work. You decide!

rxhanson:
Thanks! Good to know. I'm going to leave it with the right and left and let the user figure out that it has different functionality on vertical displays. I think the best solution might be to verbosely communicate the vertical display behavior in some kind of popover, but I don't have any desire to add the complexity to the code, UI, and translations any time soon. Thanks for the other PR as well!

PR #50, opened 2019-12-11
Updated Korean Main.strings

Zeta611:
Replaced Google-translated strings with more natural wordings.
I don't think there are proper translations for "two thirds," "Last Third," etc., so I replaced them with 1/3 or 2/3 with a translation for "left" and "right".


Zeta611:
I updated translations for directional words, because the first third is not actually the left third but the top third as mentioned [here](https://github.com/rxhanson/Rectangle/pull/42#pullrequestreview-324750090).

rxhanson:
Thanks @Zeta611! Next release will have these updates. 

Send me a message on https://gitter.im if you're interested in a free Multitouch license.

PR #53, opened 2019-12-23
Improve German translation

noiob:


Rectangle/de.lproj/Main.strings
```diff
@@ -261,25 +261,25 @@
...
 /* Class = "NSMenuItem"; title = "Select All"; ObjectID = "Ruw-6m-B2m"; */
-"Ruw-6m-B2m.title" = "Alle auswählen";
+"Ruw-6m-B2m.title" = "Alles auswählen";
 
 /* Class = "NSMenuItem"; title = "Jump to Selection"; ObjectID = "S0p-oC-mLd"; */
 "S0p-oC-mLd.title" = "Zur Auswahl springen";
 
 /* Class = "NSWindow"; title = "Rectangle Preferences"; ObjectID = "STb-JK-oB1"; */
-"STb-JK-oB1.title" = "Rectangle Einstellungen";
+"STb-JK-oB1.title" = "Rectangle-Einstellungen";
```
rxhanson: The last German translation contribution from a native speaker had removed the hyphen but I imagine it's a nuance that's probably fine either way.

noiob: the Hyphen is a bit more right but I doubt the average person would care (here's a German source http://www.sprachschach.de/bindestrich-zwischen-eigennamen/)

Issue #69, opened 2020-01-16
Function key shortcuts aren't displayed properly from menu bar


claytonrcarter:
macOS version: 10.13.6  
Rectangle version: 0.15 (19)

Hi. I just downloaded the app and have been playing around with it. Great app; thanks for it!

If I set some shortcuts to the extended function keys on my keyboard, they don't appear to be displayed correctly in the menu that drops down from the menubar. They all just display as "F", not "F18", "F19", etc. It doesn't appear limited to the extended F keys, either; the same thing happens if I use "F5" as a shortcut. Note that they *are* displayed correctly in the preferences window.

In the preferences window:  
![Screen Shot 2020-01-16 at 9 13 36 AM](https://user-images.githubusercontent.com/1420419/72531953-85ad2200-3840-11ea-842b-fdf5baf5053e.png)

In the menu bar dropdown:  
![Screen Shot 2020-01-16 at 9 13 43 AM](https://user-images.githubusercontent.com/1420419/72531954-85ad2200-3840-11ea-941a-34d531f08b60.png)

Thanks again!

claytonrcarter:
FYI, this still exists in 0.19. (I had installed 0.15 w/ `brew`, but just upgraded via the direct download.)

rxhanson:
This got auto closed by GitHub upon merging the associated PR, but I haven't drafted a release that includes this change yet. I'll post back here when I generate a release (usually every couple of weeks or so).

claytonrcarter:
Awesome. Thank you @rxhanson and @lexmiln! No rush; the app still works
just great w/o this.

PR #71, opened 2020-01-19
Display function keys correctly in status bar menu

lexmiln:
Fixes rxhanson/Rectangle#69

`NSMenuItem.keyEquivalent` only displays the first character of the set string in the menu. To be able to display key equivalents like “F12”, we need to use font glyphs that render that entire character (which are defined by Apple in a private area of the unicode table: https://developer.apple.com/documentation/appkit/1535851-function-key_unicodes?language=objc)

MASShortcut already supports this with `MASSShortcut.keyCodeStringForKeyEquivalent`. We just need to return this property instead of the more human-readable `keyCodeString` when we want a string that can be used with `NSMenuItem.keyEquivalent`.

### DIFF
--- a/Rectangle/ShortcutManager.swift
+++ b/Rectangle/ShortcutManager.swift
@@ -37,7 +37,7 @@ class ShortcutManager {
     
     public func getKeyEquivalent(action: WindowAction) -> (String?, NSEvent.ModifierFlags)? {
         guard let masShortcut = MASShortcutBinder.shared()?.value(forKey: action.name) as? MASShortcut else { return nil }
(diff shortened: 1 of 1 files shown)


lexmiln:
image

My keyboard doesn't stretch to those high function keys but this should work up to [F35](https://developer.apple.com/documentation/appkit/1535851-function-key_unicodes/nsf35functionkey?language=objc) 😮

lexmiln:
Here's MASShortcut's docs for (keyCodeStringForKeyEquivalent)[http://cocoadocs.org/docsets/MASShortcut/2.3.6/Classes/MASShortcut.html#//api/name/keyCodeStringForKeyEquivalent].

PR #114, opened 2020-04-03
Using defaults for snapping margins

nicoladefranceschi:


Rectangle/Defaults.swift
```diff
@@ -19,6 +19,10 @@ class Defaults {
     static let almostMaximizeHeight = FloatDefault(key: "almostMaximizeHeight")
     static let almostMaximizeWidth = FloatDefault(key: "almostMaximizeWidth")
     static let gapSize = FloatDefault(key: "gapSize")
+    static let marginTop = FloatDefault(key: "marginTop", defaultValue: 5)
+    static let marginBottom = FloatDefault(key: "marginBottom", defaultValue: 5)
+    static let marginLeft = FloatDefault(key: "marginLeft", defaultValue: 5)
+    static let marginRight = FloatDefault(key: "marginRight", defaultValue: 5)
```
rxhanson, line 25: Provided "gapSize" is already here, margin could be confused for gaps between windows. Maybe "snapEdgeMarginTop", etc?

(I figure it's best to change it now so that it doesn't mess you up later if I end up changing it)

nicoladefranceschi: done!

nicoladefranceschi:
I made the required changes

rxhanson:
@nicoladefranceschi By the way, I forgot to mention that as per my contribution policy, if you want licenses for my other apps, just drop me a line on https://gitter.im

PR #122, opened 2020-04-15
"最初の1/3" is more appropriate translation for "First Third".

k1dee:
As you can see
"First Third" → "最初の1/3" not "最初の3番目"
"Center Third" → "中央の1/3"
"Last Third" → "最後の1/3"


rxhanson:
Thanks!

PR #132, opened 2020-04-25
Simplified Chinese localizations update

L1cardo:
1. correct incorrect Simplified Chinese localizations to make it more Chinese
2. fix some missing Simplified Chinese localizations

image
image

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,6 +1,6 @@
 PODS:
   - MASShortcut (2.4.0)
-  - Sparkle (1.22.0)
+  - Sparkle (1.23.0)
 
 DEPENDENCIES:
   - MASShortcut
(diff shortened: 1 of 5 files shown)


L1cardo:
> Thanks! In order to merge this in, I will need this pull request to only include Rectangle/zh-Hans.lproj/Main.strings and no other changes. Let me know if you need help getting it to that point.

I also changed one menu bar shortcut from “Q” to “cmd + Q”  you may want to check that.

About only including stuff, should I need to reopen the PR or you will help me do that?

rxhanson:
Thinking about it again, it's definitely easier if you make a new PR with just the localization, instead of trying to modify the existing one. 

Thanks for the heads up on the q modifier. I've always seen the command symbol on that menu item without explicitly setting it, but maybe there's some edge cases I'm not aware of. I'll take a look at that again.

PR #133, opened 2020-04-25
Simplified Chinese localizations update

L1cardo:
correct incorrect Simplified Chinese localizations to make it more Chinese

### DIFF
--- a/Rectangle/zh-Hans.lproj/Main.strings
+++ b/Rectangle/zh-Hans.lproj/Main.strings
@@ -3,7 +3,7 @@
 "01Z-t3-cBm.title" = "如果重复按移动窗口或设置窗口大小按键，则窗口将被移至下一个显示器显示，而不是被设置为 1/2、2/3 和 1/3 大小。";
 
 /* Class = "NSTextFieldCell"; title = "Last Two Thirds"; ObjectID = "08q-Ce-1QL"; */
-"08q-Ce-1QL.title" = "右三分之二";
+"08q-Ce-1QL.title" = "右 2/3";
 
 /* Class = "NSTextFieldCell"; title = "Top Right"; ObjectID = "0Ak-33-SM7"; */
 "0Ak-33-SM7.title" = "右上";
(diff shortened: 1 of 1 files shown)


L1cardo:
Then I will leave the shortcut issue to you

image

L1cardo:
And I think `Restore Default Shortcuts` button should be on the shortcut configuration page not on the second page like now, because this is about shortcut.
There should be a `Restore All Defaults ` button instead on the second page.

Rectangle/zh-Hans.lproj/Main.strings
```diff
@@ -132,13 +132,13 @@
...
 /* Class = "NSTextFieldCell"; title = "First Third"; ObjectID = "F12-EV-Lfz"; */
-"F12-EV-Lfz.title" = "左三分之一";
+"F12-EV-Lfz.title" = "左 1/3";
 
 /* Class = "NSMenu"; title = "Help"; ObjectID = "F2S-fz-NVQ"; */
 "F2S-fz-NVQ.title" = "帮助";
 
 /* Class = "NSTextFieldCell"; title = "Right Half"; ObjectID = "F8S-GI-LiB"; */
-"F8S-GI-LiB.title" = "右半";
+"F8S-GI-LiB.title" = "右";
```
rxhanson, line 141: I don't know Chinese, but I'm assuming that the direct translation of "half" was left in there before so that it's set apart from the "Move Right", etc (without resizing) actions? If it makes more sense to drop the half from each of these actions for Chinese, that's fine with me - just figured I'd double check and make sure that's what we want to do.

rxhanson:
The Restore Default Shortcuts button is on the second tab mainly because there's a lot to take in already on the shortcuts page, based on the feedback I've received. I'll make note of your suggestion though, in the event that things shift around in the UI and it becomes time to reevaluate.

Rectangle/zh-Hans.lproj/Main.strings
L1cardo: 右 is right and 半 is half.
But when you combine these two characters into “右半”, it would be incorrect and strange, we don’t say that in Chinese, it would be like “very thank you” instead “thank you very much”.
The proper translation of right half is “右半边” meaning “right half side”, but this is unnecessary. 
So the “right” would be fine in Chinese. 
Hope you can understand 😊

rxhanson:
Thanks! Send me a message on https://gitter.im if you're interested in a license for Multitouch or Hookshot, as per the contribution policy. I'm planning on pushing out a release of Rectangle in the next day or so that will include these changes.

PR #191, opened 2020-07-15
When moving a window to next/prev display, try to keep it in the same position if possible

jonathanjbutler:
When moving a window to next/prev display, check if it is either maximized, or left/right half of current display, if so apply that same position on new display.

Started looking into this, and these 3 positions are the ones that I use the most. This was good enough for my usage so I figured I would open a PR.

If this is a feature you would accept a PR with more work for let me know, if not no worries.

Also feels like there's a better way to implement this, so if you have any thoughts let me know.

### DIFF
--- a/Rectangle/WindowAction.swift
+++ b/Rectangle/WindowAction.swift
@@ -65,6 +65,11 @@ enum WindowAction: Int {
                          firstFourth, secondFourth, thirdFourth, lastFourth,
                          topLeftSixth, topCenterSixth, topRightSixth, bottomLeftSixth, bottomCenterSixth, bottomRightSixth
     ]
+    
+    static let positionalActions = [leftHalf, rightHalf, centerHalf, topHalf, bottomHalf,
+                                    topLeft, topRight, bottomLeft, bottomRight,
+                                    maximize
+    ]
 
     func post() {
         NotificationCenter.default.post(name: notificationName, object: ExecutionParameters(self))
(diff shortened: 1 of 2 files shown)


rxhanson:
Cool, thanks for posting this. It's something that I've thought about implementing multiple times. I'll let you know when I get a chance to think through this and give some direction, as it's functionality I would like in the app.

rxhanson:
Sorry that it took me so long to get back to you on this one. I've kinda been chewing on the best way to roll this in and haven't quite determined how I would want this to be. I would like to have the new size on the monitor be the action that put the window into it's current position before the monitor move. This way behavior would be more consistent and independent of things like windows that don't resize properly. I'm still wanting to get this feature in, and will post back when I determine a path forward.

jonathanjbutler:
Hey I rebased this and made it a bit more flexible to support a few more positions. Still somewhat hacky but I use this everyday so figured it was worth updating

rxhanson:
Thanks for the update! I think I will actually merge this for the next release with a terminal flag for enabling it. I'll post back when I have some Rectangle dev time and take another look at this.

rxhanson:
Well, I couldn't bring myself to merge it in as-is. 

I did write it in a different way for the most recent release (0.45), and if anyone is up for trying it out, that would be great.

It can be enabled with the following terminal command:
`defaults write com.knollsoft.Rectangle attemptMatchOnNextPrevDisplay -int 1`

My implementation does not compare sizes, which means that manually maximizing a window is not going to trigger this behavior. It does allow for more actions to automatically handled, though.

rxhanson:
Just doing some housekeeping and I think I'll close this one out. If you get a chance to try my implementation out or have any further thoughts on this, let me know.

PR #209, opened 2020-08-28
Add `specified`  WindowAction

oltv00:
The main purpose of the functionality is to add a hotkey to change the window size to a specific one.

The case that occurred to me is:
I work on 3440x1440 ultra wide 21:9 I want to change the screen size of some windows to a specific one.

As a basis, I took the standard screen resolution of the MacBook Pro 15', 1680x1050

You can change this with the command in the terminal:
defaults write com.knollsoft.Rectangle specifiedHeight -float 1050
defaults write com.knollsoft.Rectangle specifiedWidth -float 1680

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		12D896200936C0DE9FF41FB4 /* (null) in Frameworks */ = {isa = PBXBuildFile; };
+		30166BD024F27D6A00A38608 /* SpecifiedCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 30166BCF24F27D6A00A38608 /* SpecifiedCalculation.swift */; };
 		6327EAA4FEECF2AE087FCD30 /* Pods_RectangleTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 81D6E6EACB077B1A204F3A62 /* Pods_RectangleTests.framework */; };
 		981F27D12340E3E1006CD263 /* InternetAccessPolicy.plist in Resources */ = {isa = PBXBuildFile; fileRef = 981F27D02340E3E1006CD263 /* InternetAccessPolicy.plist */; };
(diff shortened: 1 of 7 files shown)


rxhanson:
First of all, thanks for creating a PR!

It is possible to [specify a width and height of "almost maximize"](https://github.com/rxhanson/Rectangle#almost-maximize), but it is nice to have this as an additional window action.

I am actually planning on adding another tab to the settings where a user can specify any number of additional shortcuts that correspond to a width/height/location. If I don't get to it in the near future, then I will merge this in.

MrNotch9:
@rxhanson is there an ETA for the custom dimensions feature?

rxhanson:
@MrNotch9 I did write a custom dimensions feature for Rectangle, but it just didn't fit with the Rectangle vibe. The feature inherently has some complexity to it, and because one of the app's main strong points is simplicity, it just didn't feel right adding it to Rectangle.

I did add it to [Hookshot](https://hookshot.app) in the wrench/ruler tab. (Hookshot is a paid app). 

In terms of this pull request, I've kind of had a hard time figuring out what I want to do. My current preference is to just have a single terminal command that can configure one custom shortcut and dimensions, without any modifications to the UI. This pull request adds a shortcut recorder to the UI. Parsing a shortcut from text (terminal command) would require some work, but I think that's the only way I would consider rolling it in to Rectangle.

MrNotch9:
@rxhanson I just wanted to say that I'd really love this feature in rectangle. If you're worried about simplicity, maybe you could add a toggle that enables the custom dimensions feature and another tab appears in the settings pane? If the feature's toggled off the `custom dimensions tab` would be hidden.

rxhanson:
@MrNotch9 Thanks for the feedback and the idea. I gave that some thought when I implemented it, and I still prefer keeping it out of Rectangle. Maintaining an open source app is a different kind of commitment than maintaining a closed source app, and the benefits of keeping Rectangle simple in terms of code is also a large part of my decision making on which features make the cut. As mentioned in regard to this PR, I'm not opposed to having some kind of custom functionality within Rectangle, as long as it's a slimmer, no-ui-involved change.

rxhanson:
As mentioned, I would merge a PR with this kind of functionality, minus the UI changes. In v0.41, the nsuserdefaults for shortcuts are now stored as a dictionary of the keycode and modifier flag integer values. This makes it not as big a deal (to me) to avoid adding another shortcut recorder to the UI. 

Just doing some housekeeping and closing this one out, but feel free to create a new PR without the UI changes.

Issue #171, opened 2020-06-13
gaps+distance from top


StefKors:
I would love an option to set a distance from top, which will make it possible to work together with apps like [Pecan](https://github.com/zzzeyez/Pecan)

Reading through the code it looks like an extra clause [here](https://github.com/rxhanson/Rectangle/blob/master/Rectangle/WindowManager.swift#L97-L101) might be enough. Alternatively specifying gaps as an array might also be an option.

Happy to send in a PR, but I'm not familiar with swift. If you're willing to guide me into the right direction, I could help out.

rxhanson:
I can see merging a PR with this.

Unless you're ok with smaller top half/quarter windows than bottom, having a set distance from the top would change the vertical center of the screen. This means that you would want the window calculation to be based on a `visibleFrameOfScreen` size that subtracts out your specified distance from the top. You could probably just update the `UsableScreen(s)` accordingly in the `WindowManager` before they are used for the window calculation. That way you're not doing a lot of additional calculation after the fact at the point when the gaps are applied. 

It would be best if this additional logic were pulled into a separate class, maybe as a static function so as to minimize the lines of code added to the WindowManager.

I think that should be enough to get you started?

StefKors:
Thanks @rxhanson I think it will be enough. From the sounds of it Updating the `usableScreen` size makes the most sense. Should this PR only contain options for the distance to the top, or should it also have the option for any of the other sides? 

I'll have a go at it over the weekend and open a draft PR for it.

rxhanson:
It would be best to go ahead and knock out options for other sides as well, but I would still merge the PR even if it only had top. Other sides are useful to have (particularly the bottom, but sides as well) because 3rd party dock applications aren't currently accounted for by Rectangle. Each side would have its own config value added to the `Defaults` class.

Middlerun:
I'd love to be able to specify a bottom gap. I use uBar as a dock replacement and currently I'm having to maximise windows manually so they won't underlap uBar.

mattCameron09:
@StefKors Any update on the draft for this? I also use uBar and am having issues with overlapping. Ideally I'd like to have uBar on the bottom.

StefKors:
@mattCameron09 sadly I haven't gotten around to work on this, I keep reminding myself that I should, but so far i haven't been able. If anyone else want to take a stab at it please go ahead 👍

Middlerun:
[...]

PR #220, opened 2020-09-11
Add configurable screen edge gaps

Middlerun:
Addresses issue https://github.com/rxhanson/Rectangle/issues/171

As discussed in the issue, the `WindowManager` adjusts `usableScreens.visibleFrameOfCurrentScreen` by invoking a function `ScreenEdgeGap.adjustVisibleFrame()` to reduce the bounds according to the configured gap sizes. The window calculations then use this value.

There are a few calculations that use the visible frames of the next/previous screens. In these cases they can't use `visibleFrameOfCurrentScreen`, so I've updated them to make their own calls to `ScreenEdgeGap.adjustVisibleFrame()`. I'm not super happy with that solution because I was hoping to keep that adjustment in one place for maintainability.

Another possibility would be to add a dictionary of screen frames to the `UsableScreens` struct (i.e. `visibleScreenFrames: Dictionary<NSScreen, CGRect>`), as a replacement for `visibleFrameOfCurrentScreen`. Then they can be adjusted for the gaps all in one go, and referenced from wherever.


rxhanson:
At quick glance, this looks good. I'll test it out this weekend. I agree with your thoughts on the next displays actions. I'll give this a little thought as I test it out, too.

rxhanson:
I got a chance to dig through it and I'm good with merging it in.

I have some thoughts about adjusting it so that we don't have to call the function everywhere we plan to move across screens, but that would require a decent amount of refactoring, which I'll spare you :)

Thanks for contributing! Feel free to message me on https://gitter.im if you're interested in a Hookshot or Multitouch license, as per the contribution policy.

Middlerun:
Awesome! I'll message you soon.

It just occurred to me that the readme should be updated with the new config options. I didn't think to do that in the PR.

rxhanson:
I didn't mention the readme because I prefer the readme be updated after the release goes out, not as part of this PR. If you wanted to create a new PR for the readme, it'd definitely be appreciated.

PR #219, opened 2020-09-08
Updated Japanese translation

imasaru:
Hey, Ryan!

I just installed this app last night, and it has already made a big difference in my productivity. I didn't realize how much time I've lost without it!

I hope this Japanese translation update helps. Cheers! 😁

PR #221, opened 2020-09-11
Corrected Japanese translation

imasaru:
Change suggested by @ryonakano in #219.

We are full steam ahead! 🚂


rxhanson:
Awesome, thanks!

PR #226, opened 2020-09-16
First pt-PT human localization.

nebeker:
I might take another look at this, but I fixed several glaring machine translation errors and localized RectangleLauncher.


rxhanson:
Thanks! This will go into the next release that will hopefully go out this weekend.

Feel free to message me on https://gitter.im if you're interested in a Hookshot or Multitouch license, as per the contribution policy.

PR #236, opened 2020-10-05
Translate new strings to German

xsanda:
One thing I noticed when completing these translations was that `View Logs...` used three dots rather than an ellipsis (`…`) as used elsewhere in the app and in macOS. I have corrected this in my translation to German, but not anywhere else.


rxhanson:
Thanks! Good catch on the ellipsis. I'll update that across the app after merging this. 

As per the contribution policy, feel free to message me on https://gitter.im for a Hookshot or Multitouch license.

PR #244, opened 2020-10-11
French localization updated for v0.35

Yann-R:
Full proofreading of French localization: corrected existing vocabulary according to Apple guidelines and added all new strings not translated.


rxhanson:
Thank you! 

Send me a message on https://gitter.im if you want a license for Multitouch or Hookshot.

PR #248, opened 2020-10-14
Update pt-BR translations

taismassaro:
Updating the existing translations and adding the new ones from the last update!

I've been meaning to do this for some time and got the extra motivation due to the current Hacktoberfest. Would appreciate if you could label the PR `hacktoberfest-accepted` if you find that it contributes to the project! :)


rxhanson:
Awesome, thanks!

PR #253, opened 2020-10-26
Update Japanese translation

ryonakano:
- Fix wrong translations
- Follow the translations used in macOS
- Complete missing translations


rxhanson:
Thanks!

Issue #210, opened 2020-08-30
Split screen to fit 4 windows horizontally.


kamzata:
I own a LG 49WL95C-W with a 5120x1440px resolution. It would be nice fit 4 windows horizontally.

rxhanson:
I don't have plans to add this in myself, but would consider a pull request for it.

kamzata:
> I don't have plans to add this in myself, but would consider a pull request for it.

It would be great for these new size 32:9 monitors. 3 windows fit good but fit 4 can be more productive and still usable since 1 of them would be 1280 wide.

rxhanson:
I changed my mind - I'm going to add this in to the next release. I was going to add sixths and it makes sense to just add this in at the same time.

kamzata:
> I changed my mind - I'm going to add this in to the next release. I was going to add sixths and it makes sense to just add this in at the same time.

This is a great news. By the way, could I ask you... what is the relation between Rectangle and Magnet?

rxhanson:
Mind giving the beta a test? https://github.com/rxhanson/Rectangle/releases/tag/0.33

kamzata:
> Mind giving the beta a test? https://github.com/rxhanson/Rectangle/releases/tag/0.33

Glad to be your guinea pig.

Tinche:
Hah, excellent! I too am looking at getting a 49-inch ultrawide and am looking for tools to optimize my flow. I was going to suggest fourths but I see there's already work being done on the more-sizes branch.

Apart from first-forth and last-forth, could I suggest a command for the middle two fourths (middle half) as well? The idea being, on an ultrawide, the window layout being 25%-50%(main focus)-25%.

rxhanson:
@kamzata You mean dragging to screen edges? What's your expected behavior here?

@Tinche That's in the beta, if you wanted to give it a try before I get a release out the door. The beta is stable, it's mainly held up because I have to prep it for localization and I had a couple more things I was going to add in.

kamzata:
> @kamzata You mean dragging to screen edges? What's your expected behavior here?
> 
> @Tinche That's in the beta, if you wanted to give it a try before I get a release out the door. The beta is stable, it's mainly held up because I have to prep it for localization and I had a couple more things I was going to add in.

Yeah, I meant just that. My expected behavior is the same as for "3 windows" setup, but just for 4.

rxhanson:
[...]

PR #283, opened 2020-12-05
Add resizing to the moveToEdge commands

xsanda:
Follows on from https://github.com/rxhanson/Rectangle/issues/210#issuecomment-739175183

### Final version

This version adds a command-line switch `resizeOnDirectionalMove`, which, when enabled, would perform the triple-state (½, ⅔, and ⅓) cycle on directional moves. This overrides any moving between screens for these commands.

This command would respect gaps in the direction moved. For continuity this would only be when this new switch is set.

There are no changes to the UI.

### Original version (first commit), for historical reasons

> The new command would preserve the current width and horizontal position, but set the height to 50% of the screen, and the vertical position against the top edge. Press it once more, and it would move to the bottom edge. Press it once more and it would restore to the original size.
> 
> I have named this Move to Top Half, and implemented the inverse Move to Bottom Half too.

### TODO

- ~~New icons~~
- ~~Add third state of restoring the height~~
- [x] Implement changes as suggested in https://github.com/rxhanson/Rectangle/issues/210#issuecomment-739373855:
  - [x] Add a command-line switch `resizeOnDirectionalMove`
  - [x] Replace manual dual-state with a triple-state cycling through ½, ⅔, and ⅓.
    - [x] Replace vertical logic
    - [x] Replace horizontal logic
  - [x] Disable gap handling on a per-dimension basis, instead of counteracting it, but only use the new behaviour when the new switch is set.
  - [x] Revert the UI changes, as everything else can go on the command line.

### Side effects

- `WindowAction#gapsApplicable` is now a `Direction`, i.e. an OptionSet of horizontal and vertical. This allows the updated commands to respect gaps in the direction they are moving, without shrinking in the perpendicular direction.
- MoveUpCalculation and MoveDownCalculation have been merged, as they are basically the same, and MoveLeftCalculation and MoveRightCalculation had already been merged.


rxhanson:
When you're looking at implementing the three state cycling - for all the new actions that I've been adding, I've been moving away from using `RepeatedExecutionsCalculation` because it's kind of restrictive to tie a calculation to three repeated executions; it's easier to add new repeated executions if needed in the future without it. (If you already implemented it with using that protocol, that's fine, too - you don't have to go back and change it)

rxhanson:
I agree with your thoughts on the gap - to me it makes more sense to handle it in the gap calculation itself as opposed to compensating for it.

xsanda:
@rxhanson Should I update maximiseHeight to respect vertical gaps? It seems like the inverse of this command to some extent, so it seems strange that it then becomes slightly taller than all other windows

Rectangle/WindowCalculation/MoveLeftRightCalculation.swift
```diff
@@ -8,77 +8,98 @@
...
+class MoveLeftRightCalculation: WindowCalculation, RepeatedExecutionsCalculation {
     
     override func calculate(_ params: WindowCalculationParameters) -> WindowCalculationResult? {
-        let usableScreens = params.usableScreens
-        if params.action == .moveLeft {
-            return calculateLeft(params, screen: usableScreens.currentScreen)
-        } else if params.action == .moveRight {
-            return calculateRight(params, screen: usableScreens.currentScreen)
+        
+        if Defaults.resizeOnDirectionalMove.enabled {
```
rxhanson, line 20: This will ignore `Defaults.subsequentExecutionMode.traversesDisplays` when `resizeOnDirectionalMove` is enabled. I think I would prefer to have it still honor the `traversesDisplays` setting since that toggle is currently in the UI.

rxhanson (changes requested):
This looks really good! Just a couple of items to fix up before merge. 👍

rxhanson:
On the maximize height - that does make sense to me to add gaps if you don't mind adding that in as well. 

I think it's acceptable to leave off the gaps for moveToEdge commands with resizing turned off at this time. If I do get feature requests for it, then I can let you know if you'd be up for taking a look at it in the future.

Thanks!

Rectangle/WindowCalculation/LeftRightHalfCalculation.swift
```diff
@@ -70,12 +70,9 @@ class LeftRightHalfCalculation: WindowCalculation, RepeatedExecutionsCalculation
 
     func calculateLeftAcrossDisplays(_ params: WindowCalculationParameters, screen: NSScreen) -> WindowCalculationResult? {
                 
-        if let lastAction = params.lastAction, lastAction.action == .leftHalf {
-            let normalizedLastRect = AccessibilityElement.normalizeCoordinatesOf(lastAction.rect, frameOfScreen: params.usableScreens.frameOfCurrentScreen)
-            if normalizedLastRect == params.window.rect {
-                if let prevScreen = params.usableScreens.adjacentScreens?.prev {
-                    return calculateRightAcrossDisplays(params, screen: prevScreen)
-                }
+        if isRepeatedCommand(params) {
```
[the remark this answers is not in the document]
xsanda: Good spot, sorry about that. That should be fixed now.

Rectangle/WindowCalculation/MoveLeftRightCalculation.swift
xsanda: I’ve updated this behavior and the docs to reflect this.

rxhanson:
Looks good. In testing it out, I can see why you created https://github.com/rxhanson/Rectangle/issues/293 and the associated PR. Thanks for working on the move left/right behavior & across displays even though it wasn't part of your original plan.

I'm going to update the Readme after merging this, since I probably won't roll out another release until more toward the end of the week as 0.40 just got released a couple days ago.

PR #284, opened 2020-12-05
Fix gap on center half and positioning of vertical sixths

xsanda:
A couple of bug fixes:

- Add a gap around centerHalf, for consistency with all other fractional positions.
- Correct orientation of Bottom Right Sixth on vertical displays.
- Update the repetition sequence on center sixths in portrait: previously the window would jump between the two sides of the screen, whereas now it will stay on the same side as it cycles through the center-sixth and the two two-sixths positions.

Before and after (center sixths commands):

before after


rxhanson:
Looks good. Thanks!!

Issue #165, opened 2020-05-29
How to describe Mac OS bug?


DrRob:
How should I describe the "_Window moving/resizing appears animated/smooth instead of quick, and windows don't end up where you expect_" bug in the Mac OS feedback form?  Is it specific to Rectangle or more general?  If the latter, how should I describe it without reference to Rectangle?

e.g. I guess it might be something like this: "When Accessibility utilities try to resize/reposition a window, for some applications such as Google Chrome this results in the window size/position animating and not ending up in the size/position requested.  See: https://github.com/rxhanson/Rectangle#window-movingresizing-appears-animatedsmooth-instead-of-quick-and-windows-dont-end-up-where-you-expect"

It's quite a showstopper for me as it makes Rectangle useless on Google Chrome windows.

I'm using v0.25.1, but the delay is annoying.

rxhanson:
Good question. Your proposed text looks good to me. If you are aware of what's causing it - if it's any of the triggers that I've listed then it might help to add that as well.

If you wanted to reference issues where this was reported for Rectangle, you can reference these:
https://github.com/rxhanson/Rectangle/issues/94
https://github.com/rxhanson/Rectangle/issues/29
https://github.com/rxhanson/Rectangle/issues/15

I think the main thing will be that the folks Apple side will correlate all of the bugs reported for this to the same issue, and the text you have should be enough for that.

Thanks!

DrRob:
I don't know if this is relevant but I was previously using Spectacle and this problem starting happening there.  I went to check for updates and discovered that Spectacle is now unmaintained, so looked for an alternative, and that's how I came across Rectangle.

The strange thing is it's only affecting my work Macbook (a 2018 model) - my home Macbook (a late 2013 model, still using Spectacle) isn't having the problem.

I tried unticking all the other apps that have Accessibility permissions and that made no difference.

rxhanson:
Unfortunately, I'm in the dark as much as you are on what could trigger this. The fact that some users experience it when an external monitor is connected means it's maybe not necessarily directly tied to accessibility applications, but just to some level of macOS infrastructure that handles window movement. 

This happening in Spectacle (and Rectangle) is just exposing the problem, not creating it. I was able to reproduce it with Magnet as well, using both the macOS on-screen keyboard and a Russian app, Punto switcher.

[...]

Issue #94, opened 2020-02-14
Incorrect sizing and positioning with external display since new updates of macOS and Rectangle


marlonrichert:
macOS version: 10.15.3 (19D76) 
Rectangle version: 0.22 (26)

For example…

I try to Maximize by dragging to the top:
![Screen Shot 2020-02-14 at 8 51 58](https://user-images.githubusercontent.com/644725/74508308-6967d800-4f07-11ea-9993-025b3919aa2c.png)
but then this happens:
![Screen Shot 2020-02-14 at 8 52 09](https://user-images.githubusercontent.com/644725/74508324-77b5f400-4f07-11ea-9797-7585b27fb4ee.png)

I try to do Left Half by dragging to the left side of the screen:
![Screen Shot 2020-02-14 at 9 31 21](https://user-images.githubusercontent.com/644725/74510606-cf0a9300-4f0c-11ea-88bf-3305dbdcd2b1.png)
but then this happens:
![Screen Shot 2020-02-14 at 9 31 28](https://user-images.githubusercontent.com/644725/74510633-dd58af00-4f0c-11ea-876e-fef3687edf3e.png)

Similar (but different!) unpredictable things happen when using the keyboard shortcuts or the menu, and it seems to vary from app to app.

Restarting Rectangle doesn't help, but **restarting the app does.** 😮❓

Unplugging my external display also seems to fix the problem —until I plug it back in.

marlonrichert:
Also, when an app is affected by this behavior, it resizes really slowly.

Apps that behave correctly resize instantly.

rxhanson:
This appears to be an odd manifestation of this issue: https://github.com/rxhanson/Rectangle/issues/29, which I haven't solved and had chalked it up as something that most likely will have to be fixed Apple side.

Did you recently install any other apps that required you to check the box in System Prefs -> Security & Privacy -> Privacy -> Accessibility? 

The main thing that appears to be different for your issue is that the external monitor affects it.

What's the make/model of your external monitor? I'll pass this to Apple when I create a bug report for them.

marlonrichert:
I have an Asus PB278.

sic2:
Same issue here. Interestingly, I experienced this first with spectacle and then switched to rectangle in the hope that it was fixed but....no no

rxhanson:
@sic2 That's because it's a macOS bug. It is actually a big help if anyone that has this issue can file a bug with Apple - the more people that file it, the more likely it will be fixed.

[...]

Issue #29, opened 2019-11-15
Broken resize/reposition behaviour sometimes, on some apps


dylan-chong:
With some apps, sometime after rebooting the machine, resizing does not resize a window to its proper size.

On the screen recording, I am trying to resize the window to the left half by repeatedly pressing command+alt+left. As you can see, the window does not get repositioned/resized properly

![ezgif-4-5e6caee4f299](https://user-images.githubusercontent.com/17482349/68982039-53943980-086a-11ea-8eec-f162bfc8b23f.gif)

I am using Mac OS 10.14.6  on a 13 inch MacBook Pro, with a 1440p display. The issue happens on any size display

A similar issue also happens with spectacle

rxhanson:
I added a [Troubleshooting section](https://github.com/rxhanson/Rectangle#troubleshooting) to the readme. Most importantly there, you'll need to enable logging and add some logs to this issue.

Which apps are affected? 
Do some apps move/resize properly? 
Does the issue resolve itself at some point and start working properly?
Do the windows move/resize properly for the initial time period after a reboot?
Is there anything that triggers the issue, like starting an application or adding/removing displays?
How long does it work properly before the issue is seen?

dylan-chong:
I have just realised that this seems to work the first time I open an app (e.g. with Spotlight), but if I switch apps then switch back (commands tab twice), then the issue starts to happen. 

This affects notes, Safari, finder, Spotify, Kitty basically anything as far as I can tell.

There doesn't appear to be any debug messages with the filter `process:Rectangle any:calculatedRect`, however removing the `any:calculatedRect` filter shows the following logs when I press cmd+alt+left

```
debug	13:24:25.283306 +1300	Rectangle	Getting process information for [ 0x0/0x6a06a]  from shared memory failed, so asking server for it
debug	13:24:25.283364 +1300	Rectangle	sending={asn=434282U, restrictto=["Hidden","LSApplicationHasSignalledItIsReady","LSDisplayName","ApplicationType","pid","ChangeCount"], command=360, }
debug	13:24:25.283780 +1300	Rectangle	 reply={result={ApplicationType="Foreground", LSApplicationHasSignalledItIsReady=true, ChangeCount=530, pid=926, Hidden=false, LSDisplayName="Console", }, success=true, cacheable=false, }
debug	13:24:25.924970 +1300	Rectangle	Getting process information for [ 0x0/0x6a06a]  from shared memory failed, so asking server for it
[...]

Issue #15, opened 2019-09-27
Maximize function does not resize to fullscreen in certain cases


preetsethi:
Depending on window placement (and possibly size of the window?), when I trigger the _**Maximize**_ function, there are certain cases where it resizes or moves the window, but does not resize the window to the maximum size allowed on the first call. In the end, I am able to chain together a few _**Maximize**_ calls to get the desired effect, usually requiring 3 calls (or pressing the hotkey 3 times in a row).

Here's some screenshots to demonstrate what I am talking about.

1. Starting window placement (I used Rectangle's _**Bottom Left**_)
<img width="1680" alt="Screen Shot 2019-09-27 at 6 39 09 AM" src="https://user-images.githubusercontent.com/8645918/65774385-fb0ab080-e0f2-11e9-832f-ce73a6d1c038.png">

2. After I trigger _**Maximize**_ (first invocation):
<img width="1680" alt="Screen Shot 2019-09-27 at 6 39 22 AM" src="https://user-images.githubusercontent.com/8645918/65774308-d1ea2000-e0f2-11e9-8f93-1a226ccfeb0a.png">

3. After I trigger _**Maximize**_ (second time):
<img width="1680" alt="Screen Shot 2019-09-27 at 6 39 35 AM" src="https://user-images.githubusercontent.com/8645918/65774361-ef1eee80-e0f2-11e9-9dc4-ef934c699137.png">

4. After I trigger _**Maximize**_ (third time) I finally get the desired result:
<img width="1680" alt="Screen Shot 2019-09-27 at 6 39 44 AM" src="https://user-images.githubusercontent.com/8645918/65774464-27263180-e0f3-11e9-9320-ac638568ed2b.png">

As I said, this only happens in certain cases. If I use _**Top Left**_ for some window, and then I try to _**Maximize**_, I get the desired effect on the first call.

These are the only cases I can confirm are 100% reproducible for me. However, I see the issue quite a bit with various window placements / sizes. When I start the window in the _**Bottom Right**_ placement, I get annoyingly variable results when triggering _**Maximize**_ next. Sometimes it will succeed on the first call but more often it will require two calls.

I am not running any external monitors, I experience these issues only using the built-in display on my Macbook Pro. I don't run multi-monitor at all these days, but I can easily set this up and test under that setup as well. 

So to provide some more background (and probably useless info):
[...]

PR #285, opened 2020-12-06
Fix incorrect window resizing by disabling AXEnhancedUserInterface

onebytegone:
This PR should fix the issue noted in the README regarding [windows not resizing to the proper dimensions](https://github.com/rxhanson/Rectangle/blob/cd87fbfb98c06f7a75661a04f1223bac95972cd4/README.md#window-movingresizing-appears-animatedsmooth-instead-of-quick-and-windows-dont-end-up-where-you-expect).

### To reproduce the original issue:

1. Open ["Accessibility Inspector"](https://developer.apple.com/library/archive/documentation/Accessibility/Conceptual/AccessibilityMacOSX/OSXAXTestingApps.html)
1. Inspect the window of an app know to have this issue (e.g. Google Chrome, Sublime Text, Outlook)
1. Select the application via the "Hierarchy" section
1. Note the value of "Enhanced User Interface", changing it to true if needed
    enhanced-ui-attr
1. Trigger a resize of the window using Rectangle

Using the above process, Rectangle v0.39 will not resize the window properly when "Enhanced User Interface" is enabled. With changes in this PR, Rectangle will resize the window correctly, regardless of the original value of "Enhanced User Interface".

### Tested on:

* macOS 10.15.7

Refs: #15, #29, #94, #165

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -10,6 +10,13 @@ import Foundation
 import Carbon
 import Cocoa
 
+// The AXEnhancedUserInterface attribute is undocumented. However, it does appear in the
+// Accessibility Inspector [1] under the name "Enhanced User Interface" and in the list of
+// attribute names returned by AXUIElementCopyAttributeNames for an AXUIElement with a
+// role of kAXApplicationRole.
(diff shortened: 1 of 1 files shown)


rxhanson:
This is awesome. I spent so much time trying to figure this out. Thanks so much!!!

PR #297, opened 2020-12-13
Factor out logic of cycling 1/2 - 2/3 - 1/3

xsanda:
I refactored this as part of #283, but figured it was probably too off-topic for the same PR, so I held off submitting it until that was merged. This extracts the logic for handling the 1/2 - 2/3 - 1/3 to a sub-protocol of `RepeatedExecutionsCalculation`, called `RepeatedExecutionsInThirdsCalculation`.

I’ve now realised that `RepeatedExecutionsCalculation` doesn’t have any members that aren’t `RepeatedExecutionsInThirdsCalculation`, but I imagine that may change in future so I’ve left it in. It may be worth replacing `RepeatedExecutionCalculation`’s named methods with an array of function at some point, if the number of steps in the cycle needs to vary.

### DIFF
--- a/Rectangle/WindowCalculation/MoveUpDownCalculation.swift
+++ b/Rectangle/WindowCalculation/MoveUpDownCalculation.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-class MoveUpDownCalculation: WindowCalculation, RepeatedExecutionsCalculation {
+class MoveUpDownCalculation: WindowCalculation, RepeatedExecutionsInThirdsCalculation {
     
     override func calculateRect(_ params: RectCalculationParameters) -> RectResult {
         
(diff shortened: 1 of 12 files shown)


Rectangle/WindowCalculation/MoveUpDownCalculation.swift
```diff
@@ -35,16 +35,8 @@ class MoveUpDownCalculation: WindowCalculation, RepeatedExecutionsCalculation {
...
-    }
-    
-    func calculateSecondRect(_ params: RectCalculationParameters) -> RectResult {
-        return calculateGenericRect(params, fraction: 2 / 3.0)
-    }
-    
-    func calculateThirdRect(_ params: RectCalculationParameters) -> RectResult {
-        return calculateGenericRect(params, fraction: 1 / 3.0)
+    func calculateFractionalRect(_ params: RectCalculationParameters, fraction: Float) -> RectResult {
+        return calculateGenericRect(params, fraction: fraction)
```
xsanda, line 39: This one was slightly frustrating, as `calculateFractionalRect(_ params: RectCalculationParameters, fraction: Float?)` isn’t a valid implementation of the protocol’s `calculateFractionalRect(_ params: RectCalculationParameters, fraction: Float)`, so I had to separate into `calculateFractionalRect` and `calculateGenericRect`.

rxhanson: Right, that's not a big deal to me to have it call calculateGenericRect.

rxhanson:
Thanks for refactoring this. Definitely a lot smoother to work with now.

Issue #13, opened 2019-09-12
Support for "next third"


daxmc99:
Hello! 
Thanks for working on this project!
One feature that I miss from Spectacle is the "Next Third" option. 
I would love to see that in the next release. 

I'm new to swift but I will see if I can't figure out how to port this over.

rxhanson:
@daxmc99 @drpebcak , this functionality is now included in https://github.com/rxhanson/Rectangle/releases/tag/v0.9. 

It's actually just built in to the first third and last third actions. So, if you repeatedly execute first third, it will cycle through thirds (first, center, last). If you repeatedly execute last third, it will cycle through thirds in reverse order (last, center, first).

goldcaddy77:
Finding this years later... nice compromise, works great!

PR #333, opened 2021-01-18
Update README.md with third cycling

plourenco:
Hello 👋

First of all, thank you for giving continuity to such an essential tool.
I'm part of the group moving from Spectacle and one of the features I missed the most was **third cycling**. I understand how granular third shortcuts can be helpful, however, it can also become a burden having to memorize multiple shortcuts.

Luckily I found #13 with a solution on how to emulate third cycling. Since this isn't straightforward, I would like to suggest clarifying that in the differences for future pre-Spectacle adopters like me.


rxhanson:
Thanks!

PR #367, opened 2021-02-03
Update halve-to-thirds setting label in README

davidjb:
The option `Cycle displays` isn't present in the preferences window any longer since the UI label changed.  This PR updates the readme to reflect the new name.


rxhanson:
Thanks!

PR #378, opened 2021-02-11
Change label for hidden shortcuts

ajs256:
The "..." label doesn't well describe what is hidden by the control. I changed it to `More...`. Note that this is untested because I can't get it to build, but it looks okay in the storyboard.

image


rxhanson:
Hi, thanks for looking to contribute!

The lack of text here is intentional. I try to avoid additional text where ever I can so that it minimizes the amount of churn for existing localization translations and makes for one less thing that must be translated and understood by a translator for new localization translations. The ellipsis by itself is pretty universal, and I expect the action of showing more shortcuts is immediately learned after one click if it is originally unclear. 

A concerning issue is that you cannot get the project to build. Did you install the CocoaPods? 
https://github.com/rxhanson/Rectangle#running-the-app-in-xcode-for-developers

If so, where did you get stuck?

I'm going to close this one out (sorry!) since the desired outcome is the current implementation.

ajs256:
> Did you install the CocoaPods?

Ooh, thanks.

PR #394, opened 2021-02-28
Repeated commands move window to same position on next screen

kitschpatrol:
Following up on discussion https://github.com/rxhanson/Rectangle/discussions/374

This is a little kludgy... there's a recursive call to `execute`, and a white lie gets passed to `recordAction`. Definitely open to improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,27 +54,64 @@ If you wish to change the default shortcuts after first launch, use the followin
 defaults write com.knollsoft.Rectangle alternateDefaultShortcuts -bool true
 ```
 
-### Halves to thirds (repeated execution of half and quarter actions)
-Halves to thirds is controlled by the `Move to adjacent display on repeated left or right commands` setting in the preferences. 
-If this setting is not checked, then each time you execute a half or quarter action, the width of the window will cycle through the following sizes: 1/2 -> 2/3 -> 1/3.
+### Repeated Command Behavior
 
-The cycling behavior can be disabled entirely with:
+Rectangle implements a number of behaviors for handling repeated commands, only some of which are exposed in the preferences.
+
+The cycling behavior can be set with
+
+```bash
+defaults write com.knollsoft.Rectangle subsequentExecutionMode -int <VALUE>
+```
+
+Where `<VALUE>` is one of the following:
+
+#### `0`: Halves to thirds (Spectacle behavior)
+
+Each time you execute a half or quarter action, the width of the window will cycle through the following sizes: 1/2 → 2/3 → 1/3.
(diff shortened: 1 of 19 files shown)


rxhanson:
Sorry this has taken me a while to get around to this. My plan is to go through it this weekend.

rxhanson:
It ended up making more sense to me to write this in a little different way:
https://github.com/rxhanson/Rectangle/compare/alt-cycle-displays

There's good and bad things about both ways, but the way I wrote will contain the cycleMonitor execution mode to just one check (and the additional switch case check in the leftRight calculation). It also removes the recursive call and messing with recordAction.

I tested out my branch and it works for me. Before I merge it in, can you test it out, too? Also let me know if you have any feedback on my approach.

rxhanson:
I just went ahead and merged in my branch. Planning on doing a release soon, so if you want you can just wait for the release and report back how it's working.

Closing this one out.

kitschpatrol:
Thanks Ryan, sounds good, will test the next release.

kitschpatrol:
Hi Ryan thanks again for including this feature.

I just tested the v0.44 release, and there's just one difference in the behavior of your implementation vs. my PR — basically, if a window is already in the target position (and was placed in this position manually), then you have to invoke the Rectangle command twice to get it to jump to the next display.

**To reproduce:**

1. On a 2+ display Mac, open a new window and make it maximized manually.
2. Invoke Rectangle's maximize command.

**Expected behavior:**
The window moves (maximized) to the next display.

**Observed behavior:**
Nothing happens.


The _second_ invocation of the maximized command works, I think because the original command is in Rectangle's history at that point and so is detected as a repeat. This is how I ended up with the recursive call to `execute` with the fake history entry, but maybe there's a better way to handle this and still get the expected behavior.

This applies to any command where the window's size already matches Rectangle's target size— maximized is just a convenient example. (I run into this quite a lot in v0.44 since many apps open their windows "maximized" by default.)

Happy to spend more time on this, but would be a few weeks from now.

rxhanson:
Right, that makes sense. Sounds like it should be a quick fix - just checking whether or not the window takes up the entire visible frame of screen in the logic specific to subsequentExecutionMode 4 (in the ShortcutManager if my memory serves me right). It doesn't sound like we need to be concerned with any other window sizing, like halves, etc. I still prefer going the existing route and avoiding both the recursive call and modifying the WindowManager.

I'm in no rush with this one, so I'll let you take a look at it when you have time. That way I won't be implementing something further based on my assumption with only measuring for "maximized dimensions" of the window if that's not really what you're after.

rxhanson:
I went ahead and added in the logic to check window size, just for maximize. If you want to add in more window size comparisons, I don't mind merging in any PRs for that.

PR #410, opened 2021-03-10
Update.

nebeker:


nebeker:
Some strings were reverted with the addition of semicolons. I hadn't noticed it before.

rxhanson:
Thanks! It'll be in the next release.

PR #478, opened 2021-05-29
Add `specified` WindowAction

oltv00:
Depends on #209, I made changes to my code.

For now, to add the shortcut to the `specified` action, you should write some defaults like:

```sh
defaults write com.knollsoft.Rectangle specified -dict-add keyCode -float 8 modifierFlags -float 1966080
```

because I didn't change any UI and didn't add the menu to add shortcuts

keyCode: 8 
modifierFlags: 1966080
it's `CMD + OPTION + CTRL + SHIFT + C`

I have no idea how to make it more user-friendly without changes in UI. :)

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		12D896200936C0DE9FF41FB4 /* (null) in Frameworks */ = {isa = PBXBuildFile; };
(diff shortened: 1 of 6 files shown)


rxhanson:
Thanks! I'll probably roll out a release in the next week or so. 

Side note: An easy way to get keycodes & modifier flags for shortcuts outside of looking at Rectangle defaults is with the free key codes app: https://apps.apple.com/us/app/key-codes/id414568915?mt=12

Issue #225, opened 2020-09-16
Add spacing option


saschaeggi:
I love Rectangle and it replaced Magnet for me.

I know you're not accepting any new feature request but maybe you still have an ear for this idea: it would be very cool to set an offset in the settings (spacing) for Windows. so that Windows don't go edge-to-edge but would have a spacing between them. See example:

Without spacing:
<img width="3360" alt="without_spacing" src="https://user-images.githubusercontent.com/5558166/93304884-053f2680-f7fe-11ea-9d0e-f47ca15574c7.png">

With spacing option:
<img width="3360" alt="with_spacing" src="https://user-images.githubusercontent.com/5558166/93304896-0a9c7100-f7fe-11ea-8dd0-24bf2e1f2c17.png">

This would be a great enhancement setting.

Keep up the good work & Thanks for considering,
Sascha

rxhanson:
https://github.com/rxhanson/Rectangle#adding-gaps-between-windows

saschaeggi:
How could I oversee this? So awesome! 
Maybe we can have this as a setting in the preferences as well? 💪💙😍

rxhanson:
I think I will add this in as a preferences setting in a future release. I have avoided it so far to keep the prefs simple (and avoid yet another thing that has to get localized).

PR #492, opened 2021-06-15
Re-add documentation on adding window gaps

manojkarthick:
I was looking to add window gaps and noticed that the feature is available through this [issue](https://github.com/rxhanson/Rectangle/issues/225). Looks like the documentation on how to enable it was removed in this [commit](https://github.com/rxhanson/Rectangle/commit/3e9d5f12c1366257203e7eb832fcd1ab299573be#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5). 

I wasn't sure if that was removed inadvertently, so I figured I would open a PR to add it back. If you made the explicit choice to remove it from the README, please let me know and I will close the PR.

Thanks a lot for creating Rectangle!

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,14 @@ defaults write com.knollsoft.Rectangle screenEdgeGapLeft -int 10
 defaults write com.knollsoft.Rectangle screenEdgeGapRight -int 10
 ```
 
+### Setting gaps between windows
+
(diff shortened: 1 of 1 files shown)


rxhanson:
Hey, thanks for looking to contribute!

This was removed because you can configure it in the settings tab of the preferences window. 

Apologies that I didn't make it clear in that commit to the readme.

PR #500, opened 2021-06-26
Add window calculations for three fourths of the screen.

bjornorri:
I added the option to snap to the left and right 3/4 of the screen, along with the relevant menu items.

The motivation is to complement the 1/4 option already present in the app. I use an ultra-wide monitor with a 21:9 aspect ratio. If I set the window width to 3/4 of the screen, I get a window that is approximately the size of a "normal" 16:9 monitor. I really like this split e.g. for working in a code editor with a 1/4 sized terminal on the side.


rxhanson:
Thanks for looking to contribute!

This all looks fine to me, but there are a couple of things that would make sense to have done before merging. 

1. The localizable strings should be added in for each of the localized files, because typically translators are just looking to translate existing strings and won't be generating the strings themselves (as in, these typically just won't get translated if they aren't there). The English titles are fine - you don't have to translate anything yourself.
2. Can you refactor 2nd fourth and 3rd fourth repeated actions to use your calculations instead of the ones that are in those calculation files?

bjornorri:
Thanks for the feedback!

Here's my update:

1. Added empty string entries for the two new labels in each file.
2. Deleted the inline 3/4 calculation classes in the 2nd and 3rd fourth calculations and replaced the usages with references to the new calculations.

Let me know if you think of any more improvements 🙂

rxhanson:
Great, thanks! The only thing that I have left is that I kind of prefer having these come after the Last Fourth item in both the menu and the prefs window. If you didn't feel inclined to make this change in this PR, I will probably just do it myself after merging (since it's mainly just me being picky). I'll wait to merge in case you wanted to make the change. If not, let me know and I'll go ahead and merge.

bjornorri:
I can take care of it later today (I'm on CEST time). And by the way, I can totally sympathise with nitpicks myself 😄

I actually prefer this order as well, but I was trying to be consistent with how the thirds are ordered. I'll leave it up to you to decide if they need to be re-ordered as well. I tried to maintain a consistent order in arrays, translation files, etc. and the menus, so I will re-order those as well.

rxhanson:
Sounds great, thanks. No need to reorder the translation files - those are not necessarily in a consistent order to start with.

bjornorri:
Done 👍

rxhanson:
Thanks! I will most likely push out a release in the next week or so.

PR #508, opened 2021-07-05
Update Japanese locale

hiro0218:
Hi

> Rectangle v0.48
> Added the ability to move the cursor along with the window when moving across displays, along with a toggle in the preferences window.

translated it into Japanese.

PR #510, opened 2021-07-06
Update Japanese translation

ryonakano:


ryonakano:
Ah, didn't see it's already proposed. Commenting in #508 instead

Issue #157, opened 2020-05-21
Missing some translation and minor language issue (IT)


ildede:
macOS version: 10.15.4
Rectangle version: 0.29 (34)

I've just installed Rectangle to give it a try.
I'm writing this issue because i don't have the time, right now, to fix it.

This part is left untranslated:
<img width="295" alt="Schermata 2020-05-21 alle 15 08 15" src="https://user-images.githubusercontent.com/3754465/82563146-cc0be800-9b76-11ea-8ad9-4c767ad0c57e.png">

This button has text cutoff
![Schermata 2020-05-21 alle 15 07 34](https://user-images.githubusercontent.com/3754465/82563220-ef369780-9b76-11ea-8ceb-d5d16617759d.png)

The word "shortcut" is been translated as "scorciatoia" like in the images below, but Apple use the word "abbreviazioni" in system's preferences, I think is better to keep consistency with Apple convention.
<img width="497" alt="Schermata 2020-05-21 alle 15 09 26" src="https://user-images.githubusercontent.com/3754465/82563386-489ec680-9b77-11ea-905a-35253a7252d8.png">
<img width="749" alt="Schermata 2020-05-21 alle 15 09 03" src="https://user-images.githubusercontent.com/3754465/82563379-450b3f80-9b77-11ea-90fc-3ef9779515ec.png">

rxhanson:
Thanks for filing this!

1. Welcome window (untranslated): This was added after Rectangle was translated. If you feel inclined to contribute and would like to update the translations for this, that would be awesome. You can find the translation file here: https://github.com/rxhanson/Rectangle/blob/master/Rectangle/it.lproj/Main.strings
1. Authorize window: To me, the button seems ok. Am I missing something?
1. Gear tab: I went ahead and changed it. It'll go into the next release.
1. Shortcuts tab: This is actually from the open source library that I use, [MASShortcut](https://github.com/shpakovski/MASShortcut). It would be best to get it changed there, so that Rectangle doesn't diverge, and so that MASShortcut is consistent for everyone. I could file a PR for it, but maybe it would best come from an Italian speaker?

ildede:
I will definetly submit a PR in the next days ;)

To let you know about "2 Authorize window", the word _apert_ doesn't mean nothing in Italian.
It looks like a truncated word, but looking at Main.strings it's just a wrong word put there.
[...]

PR #511, opened 2021-07-06
Update Italian locale

antiasi:
Hello, I would like to help you with the translation of Rectangle! I translated the missing lines, and also changed "abbreviazioni -> scorciatoie", shortcuts is closer to italian "scorciatoie", "abbreviazioni" is when you write for example Mr. instead of Mister, it's something else. I turned "check for updates automatically" in the shorter "aggiornamenti automatici" because the translated line in the app UI was overlapping with the "check for updates" button, it was too long. I hope you may find this changes useful, thank you for your great work with this app.

### DIFF
--- a/Rectangle/it.lproj/Main.strings
+++ b/Rectangle/it.lproj/Main.strings
@@ -1,6 +1,6 @@
 
 /* Class = "NSTextFieldCell"; title = "If you repeat move or half actions, the window will move to the next display instead of 1/2, 2/3, and 1/3 size."; ObjectID = "01Z-t3-cBm"; */
-"01Z-t3-cBm.title" = "Questo vale solo per le azioni di spostamento o i mezzi spostamenti. I mezzi spostamenti non eseguono il comportamento a 2/3 e 1/3 se abilitato.";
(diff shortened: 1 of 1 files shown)


rxhanson:
Thanks for the thorough update!

On this note, I think that the shortcut recorder still has the incorrect word for shortcuts since MASShortcut didn't push a cocoapods release with the update last year (see https://github.com/rxhanson/Rectangle/issues/157). I'll look into a MASShortcut update later when I have the time - I've been considering migrating to a different shortcuts library written in Swift which is why I've been putting off thinking about this.

antiasi:
So the word "abbreviazione/i" instead of "scorciatoia/e" was choosen to be consistent with Apple translation, I can change the lines back if you want to, but in my opinion (whatever Apple says) shortcut is "scorciatoia" and this is the correct translation, "abbreviazione" is abbreviation, both in english and in italian the meanings are different.

rxhanson:
I'm good with leaving it consistent with Apple. In my opinion it is better to have consistency with Apple than with MASShortcut, especially since the path ahead will be to either update MASShortcut (or just its translation) or to replace MASShortcut with https://github.com/sindresorhus/KeyboardShortcuts.

antiasi:
ok, I'll update the file then.

PR #512, opened 2021-07-07
Update Italian locale v2

antiasi:
as we said, I updated the lines to be more consistent with Apple's localization. I changed some menu bar lines too to match the translation from system apps.


rxhanson:
Thanks! I really appreciate it. If you are interested in a Multitouch or Hookshot license, just send me an email. (as noted in https://github.com/rxhanson/Rectangle#contributing)

PR #513, opened 2021-07-12
Switch to a dropdown that allows users to choose the subsequent execution mode

nevir:
This exposes all of the subsequent execution modes via the configuration UI:

image

One thing I'm unsure about is how to go about localizing this (or whether I should)

### DIFF
--- a/Rectangle/Base.lproj/Main.storyboard
+++ b/Rectangle/Base.lproj/Main.storyboard
@@ -761,7 +761,7 @@
                     <window key="window" title="Rectangle Preferences" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="" animationBehavior="default" titleVisibility="hidden" id="STb-JK-oB1">
                         <windowStyleMask key="styleMask" titled="YES" closable="YES" miniaturizable="YES"/>
                         <rect key="contentRect" x="245" y="301" width="433" height="270"/>
-                        <rect key="screenRect" x="0.0" y="0.0" width="1680" height="1027"/>
+                        <rect key="screenRect" x="0.0" y="0.0" width="1680" height="1025"/>
                         <connections>
                             <outlet property="delegate" destination="fmq-Go-Xew" id="Jj5-zM-epV"/>
                         </connections>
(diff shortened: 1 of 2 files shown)


Rectangle/Base.lproj/Main.storyboard
```diff
@@ -3150,55 +3150,71 @@
...
+                                                    <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
+                                                    <font key="font" metaFont="menu"/>
+                                                    <menu key="menu" id="TkU-PT-5p8">
+                                                        <items>
+                                                            <menuItem title="do nothing" state="on" tag="2" id="jww-Ju-S3d"/>
+                                                            <menuItem title="cycle through displays" tag="4" id="XlM-ch-cLG"/>
+                                                            <menuItem title="cycle ½, ⅔, and ⅓ on half actions" id="gHH-BV-5kP"/>
+                                                            <menuItem title="move to adjacent display on left or right" tag="1" id="Z9d-Rl-RVq"/>
+                                                            <menuItem title="move to adjacent on left/right, or cycle size on half" tag="3" id="3GE-la-fAZ"/>
```
nevir, line 3180: Note that the tag property of these menu items indicates the value of the `SubsequentExecutionMode`

rxhanson:
Thanks! This looks good to me at first glance. I'll test it out and follow up.

I would prefer that localization placeholders are added as part of this PR, and then we leave it up to translators to go in and modify the placeholders. Unfortunately, it is unlikely that translators will translate the additional lines unless placeholders are inserted, so that needs to be done. I'm sure there is a better way to do it, but here's what I usually do:

1. In Xcode, I go to the xcodeproj (root level element) and select the project.
2. In localizations, I add a new localization (any language). 
3. Then I copy the only the new lines that I want from the generated file, and put those in each of the localization .strings files for all localizations.
4. Then I remove the localization (and it's generated file) that was just added.

The English text in the placeholders is fine. 

If you find a better way, let me know since my way is tedious and annoying.

gothic-bum:
Since there exists an `altThirdCycle` option, perhaps there should be an additional entry for "cycle ½, ⅓, and ⅔ on half actions".

rxhanson:
@nevir apologies for taking so long to respond. I do have plans to merge this in. If you're up for some of the mentioned localization work, that would definitely help speed it along. I would also prefer that the additional entry proposed by @gothic-bum would be added in, but I would still merge it without it. It still might take a little while for me to get the next release out the door. If I don't hear anything back by the time I'm ready to make a new release, then I'll fill in the gaps.

rxhanson:
I decided to go without the additional entry proposed by @gothic-bum, since that would have carried implications to other selections and there's not an elegant way to handle it.

I went ahead and fixed some autolayout issues and added localization placeholders.

PR #523, opened 2021-07-24
Italian locale minor fixes

antiasi:
I tested my updated main.strings in my Rectangle installation and I find out small errors so I fixed them. Everything else looks fine, from the menu bar to app preferences.


rxhanson:
Thanks!

PR #526, opened 2021-07-25
Add keyboard shortcut to reverse all windows on one screen

c-harding:
As discussed in #518.

Open points/possible todos:

- [ ] Naming
- [ ] Adding UI entries if desired
- [ ] Handle vertical monitors? Would this feature be desirable?
- [x] Test with more cases
  - [x] Visible side dock
  - [x] Todo mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ defaults write com.knollsoft.Rectangle todo -int 1
 ### Enable Reverse All shortcut
 This will reverse all windows horizontally on the current monitor.
 ```bash
-defaults write com.knollsoft.Rectangle reverseAll -bool true
+defaults write com.knollsoft.Rectangle reverseAll -int 1
 ```
 The default shortcut is ⌃⌥⌘⇥, but this can be changed with the following command:
 ```bash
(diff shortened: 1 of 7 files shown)


rxhanson:
Apologies for the delay. I'm still planning to merge this in for the next release. One of the things I plan on adding for the next release is a list view of shortcuts as I have in Hookshot. This will make it so that it makes sense to make a UI shortcut for this. It will still take me a little while to get around to it, as I'm currently working through an injury. I'll keep you posted when I come back to this.

rxhanson:
Man, sorry about taking so long to get back to this. I'm finally working through my injury and am starting back slowly at the keyboard again. 

I'm good with this, but there is one main thing I would change, and that's to just roll it in as a normal WindowAction, as was done with the "specified" action in this PR:

https://github.com/rxhanson/Rectangle/pull/478/files

This means that we would just remove the Defaults key for enabling this action, along with the static functions for deciphering a MASShortcut. The downside is that there wouldn't be a default value for this shortcut, and the user would have to configure it themselves provided it is not in the UI. 

When you get a chance, let me know and if you don't have a chance to get to it soon then I will just make the change and roll it in.

PR #594, opened 2021-10-07
Add brew monthly download badge

sh-cho:
image

made with https://badgen.net/homebrew


rxhanson:
Cool, thanks! I didn't know about that one.

PR #628, opened 2021-11-06
Update simplified Chinese translation.

adcros:
1. Update Todo Mode related transitions.
2. Adjust some translations for better understanding when screen in portrait mode.


rxhanson:
Thanks!

PR #652, opened 2021-11-22
Update korean translation of Todo mode

sh-cho:
:)

### DIFF
--- a/Rectangle/ko.lproj/Main.strings
+++ b/Rectangle/ko.lproj/Main.strings
@@ -672,38 +672,38 @@
 "Cancel" = "취소";
 
 /* Class = "NSButtonCell"; title = "Show Todo Mode in menu"; ObjectID = "7yS-wj-uWD"; */
-"7yS-wj-uWD.title" = "Show Todo Mode in menu";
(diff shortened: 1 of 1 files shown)


Rectangle/ko.lproj/Main.strings
```diff
@@ -672,38 +672,38 @@
...
+"6e0-ji-qXw.title" = "Todo 어플리케이션 폭";
 
 /* Class = "NSTextFieldCell"; title = "Reflow Todo shortcut"; ObjectID = "Fx0-sm-DrT"; */
-"Fx0-sm-DrT.title" = "Reflow Todo shortcut";
+"Fx0-sm-DrT.title" = "Todo 재배치 단축키";
 
-"Use frontmost.app as Todo App" = "Use frontmost.app as Todo App";
-"Enable Todo Mode" = "Enable Todo Mode";
-"Reflow Todo" = "Reflow Todo";
+"Use frontmost.app as Todo App" = "frontmost.app을 Todo 앱으로 사용";
```
sh-cho, line 683: There is small problem that object particles(을/를) should be placed based on final consonant, but it's not a big deal and I actually have no idea how to fix on Swift.

rxhanson:
Thanks!!

PR #662, opened 2021-11-24
Added option to not apply gaps for maximize

Aaron-Rumpler:
Added `applyGapsToMaximize` and`applyGapsToMaximizeHeight` preferences (on by default) which specify whether gaps are applied to 'Maximize' and 'Maximize Height' actions.


rxhanson:
Thanks!!

PR #666, opened 2021-11-27
Updated italian translation

antiasi:
I translated the new lines in Rectangle 0.49, I would like to ask you about the "Todo mode" translation because Todo in italian it's not very self-explanatory, maybe something like "In Evidenza" would be better in the italian translation, which explain the idea of a pinned app on the right of the screen.

### DIFF
--- a/Rectangle/it.lproj/Main.strings
+++ b/Rectangle/it.lproj/Main.strings
@@ -663,23 +663,23 @@
 "Cancel" = "Annulla";
 
 /* Class = "NSButtonCell"; title = "Show Todo Mode in menu"; ObjectID = "7yS-wj-uWD"; */
-"7yS-wj-uWD.title" = "Mostra modalità Todo nel menu";
+"7yS-wj-uWD.title" = "Mostra modalità In Evidenza nel menu";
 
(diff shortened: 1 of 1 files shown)


rxhanson:
I'm all for changing the name for Italian, but would it cause confusion if anyone were looking for information about Todo mode in the repo, but thinking it is instead "In Evidenza"? If you don't think that would be a problem or think it would be something that Italian users would still be able to figure out, then I'm good with the change.

GiorgioRen:
As an Italian i think "Todo" doesn't make sense, "In Evidenza" is better to explain the feature, maybe "In primo piano" ("always on top" in english) would be better?

rxhanson:
I think I'd be good with "In Evidenza", as "In primo piano" might imply that the window is actually pinned on top rather than reflowed around, since "always on top" is a common request that is not possible due to macOS restrictions. With that said, I am good with whichever translation. @antiasi let me know if you want to rework this one or just have me merge it as-is.

antiasi:
I updated the translation using In Evidenza instead of Todo, you should find the main.strings updated, tested on my local Rectangle installation and looks fine

rxhanson:
Thanks!

PR #702, opened 2022-01-02
Add ninths

slifty:
Thank you for this project!

I personally love ninths, as 4k screens are pretty common in the mac ecosystem and it's handy to be able to stack things like terminals.  I saw in [this discussion](https://github.com/rxhanson/Rectangle/discussions/396) that you would be open to a PR for adding ninths to Rectangle so I thought I'd give it a whirl.

**What's added**

1. Support for "ninths" calculations, without a UX, meaning it is possible to configure hotkeys via terminal commands for folks who are interested in that sort of thing.

-- 

Note that an earlier version of this PR included UI / UX for setting shortcuts and accessing ninths via the menu.  This was removed as part of the PR review, though the UI still exists [in this commit](https://github.com/slifty/Rectangle/commit/2b22a084017896dfeb975e7dafd9ca4e1daee4c2) if anybody is interested in that kind of thing down the line.

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -97,6 +97,16 @@
 		98FD7C5F2687BC14009E9DAF /* FirstThreeFourthsCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98FD7C5E2687BC14009E9DAF /* FirstThreeFourthsCalculation.swift */; };
 		98FD7C612687BCB6009E9DAF /* LastThreeFourthsCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 98FD7C602687BCB6009E9DAF /* LastThreeFourthsCalculation.swift */; };
 		CC0B7937429AC28C21ABF5B4 /* Pods_RectangleLauncher.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F2D8480CC730811C953FC1B6 /* Pods_RectangleLauncher.framework */; };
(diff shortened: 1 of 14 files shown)


rxhanson:
Thanks for opening the PR! For all new changes, I don't make any UI changes unless there's significant demand. Especially in this case, I only want to add 9 additional actions if there's a ton of people that want it, as I want to keep Rectangle as simple as I can. This also prevents us from having to do the additional work of localization and icons if needed.

So, I'm good with merging this in but only if all of the UI elements are reverted out of the storyboard, and the menu items are removed. To remove the menu items, you just set the `displayName` to nil in `WindowAction` for all of those new actions. Alternatively, if you really want the menu items in, then you'll need to add a flag for showing them (at which point it would also be necessary to add in icons). 

All of the new window actions added can then just be configured via terminal commands, no further code changes necessary, similar to this one: https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md#add-an-extra-centering-command-with-custom-size

slifty:
Ah yes that makes complete sense (and makes me feel glad about the two things I hadn't done yet).  I've taken the UX out of this PR (I kept it [in a branch on my fork](https://github.com/slifty/Rectangle/tree/add-ninths-with-ux) in case that UI demand crops up and it ends up being useful, though seems unlikely).

Please let me know if you'd like to see any other modifications.

slifty:
Oh I just realized: I should probably update the `TerminalCommands.md` to expose the existence of the ninths there.  I've added that in as a separate commit.

rxhanson:
Thanks! I'm planning on rolling out a release in the next couple of days.

Erik262:
Is there a way to get eights?

rxhanson:
@Erik262 You can create a PR just like this one for Rectangle or use [Rectangle Pro](https://rectangleapp.com/pro) and create custom shortcuts

PR #715, opened 2022-01-10
Fix typos in German localization

stefan-goldschmidt:
Hi! 👋  I have fixed two little typos in the german translation


rxhanson:
Thanks!

PR #723, opened 2022-01-13
Update readme files

Aaron-Rumpler:
Replaces references to Hookshot with Rectangle Pro and removes the 'Available in v0.50' notices from TerminalCommands.md


rxhanson:
Thanks!

PR #733, opened 2022-01-17
Added Swedish language

Sopor:
I'm not a macOS user so bear with me if i made some mistake. Some strings should be easier to translate with some kind of context. If you find any errors or if you think you can improve it, you are free to do so.


rxhanson:
Thanks!!

Sopor:
@rxhanson I noticed that you changed `sv-SE` to `sv`.

In Finland they speak Swedish too and to separate these two you use `sv-SE` instead of only `sv`. Finnish locale is sv-FI.

There are differences in linguistics and vocabulary between them (how much i don't know). Currency, telephone codes, and stuff like that are also different.

You do have `pt-BR` and `pt-PT`, so why not sv-SE? Doesn't macOS has sv-SE and sv-FI locales?

rxhanson:
Ah, sorry. When I added the localization, for some reason `sv` was all that popped up. Upon further inspection, there are all of the additional locales. I went ahead and changed it back to `sv-SE`. With just `sv`, I think it would have just used this localization for all the Swedish ones, which isn't that bad because it would prompt Swedish speaking users from other locales to contribute (based on other localizations I've had :). Either way, best to have it be your intended locale.

PR #740, opened 2022-02-01
Add horizontal thirds

slifty:
This PR adds four new regions which can be configured via command line.

They are "horizontal" thirds (from the perspective of landscape; in portrait they are vertical).  They are still taking up 1/3 of the screen, just... horizontally!

Specifically the four new layouts are:

```
1   2   3   4
TL  TR  BL  BR
XXO OXX OOO OOO
OOO OOO XXO OXX

Top Left (TL)
Top Right (TR)
Bottom Left (BL)
Bottom Right (BR)
```

Similar to the ninghts PR (#702) this supports repeating commands to rotate through the four options.


rxhanson:
Looks good, thanks!

Issue #713, opened 2022-01-09
Todo App gets positioned under the dock


hsingireddy:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 11.6.2
Rectangle version: 0.49 (54)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 
2022-01-08T21:11:12-05:00: Unable to obtain window id
2022-01-08T21:11:12-05:00: AX sizing proposed: (800.0, 1415.0), result: (800.0, 1415.0)
2022-01-08T21:11:12-05:00: AX position proposed: (2640.0, 0.0), result: (2640.0, 25.0)
2022-01-08T21:11:12-05:00: AX sizing proposed: (800.0, 1415.0), result: (800.0, 1415.0)

Environment:
Dock is position to the right of the screen and Reflow Todo shortcut is pressed

Expected outcome: 
TodoApp will be placed to the right of the screen next to the dock. All the other windows will be moved to the non-todo part of the screen. There will be no gap between the "other window" and "Todo App"

Actual outcome:
Todo App is placed on the right edge of the screen, under the dock. Other windows reflown are still attached to the Todo App without any gap. But, when you run shortcuts that place windows to the right of the screen (Right half, last two thirds etc.) they leave a dock sized gap between the window and the Todo App. They are leaving the correct width (Dock + todo App) but because the Todo app is under the dock, it leaves a gap.

rxhanson:
If you have window gaps configured, v0.50 should fix the gap around the Todo App window.

A macOS bug causes the windows to be placed under the dock in certain scenarios - most commonly seen when the dock is moved to a different display or unhidden. Unfortunately, there's not a good workaround for it.

hsingireddy:
Here is some more info on existing behavior with screenshots

**_-- Dock is on the left --_** 

Reflow Todo command moves the application (Safari, in this case) window under the dock. It should stop at the dock unless the window was already placed under the dock by the user.

Reflow command output ::
![Left-Dock](https://user-images.githubusercontent.com/70454212/152099449-7295a7a9-437c-448d-a1a3-999e62cb579c.png)

Other commands like Maximize will correctly place the Safari window next to the dock.
Maximize command output on the Safari Window ::
![Left-Dock-Maximize](https://user-images.githubusercontent.com/70454212/152099937-cb9acb05-1570-4c00-b8f4-00761632f813.png)


**_-- Dock is on the right --_**
[...]

PR #742, opened 2022-02-02
Bug Fix for #713

hsingireddy:
Rectangle Version 0.50 (55)
Fix for #713 

Application windows now consider the dock positition and will not be placed underneath the dock

1. todoWindow uses the adjustedVisibleFrame(which uses visibleFrame). When the dock is set to the right side, todoWindow will now placed next to it rather than under it
2. adjustedVisibleFrame is now used to calculate the positions for all non-todo application windows.
3. Changed the following in the shiftWindowOffSidebar function
	a. Uses adjustedVisibleFrame instead of screenFrame. Helps when dock is placed on the left side
	b. Handles an edge case where an application window is already placed at left edge or even further left. Origin X will be its current x in this case
	c. The condition for checking if the application window is too wide. It now considers the edge case above.


rxhanson:
Thanks!!

PR #759, opened 2022-02-16
Sort the JSON keys of the exported configuration

jparise:
This results in a more stable (and diff-able) configuration file.

`JSONEncoder.OutputFormatting.sortedKeys` was introduced in macOS
10.13, so this is protected by a runtime availability check.


rxhanson:
Thanks!

PR #761, opened 2022-02-18
Add eights without UI

JohannesTrussellRasch:
This adds the option to add Eights via the terminal. This is super useful for using an ultra-wide screen where you now can have a main window take up the center half and have 4 other windows set up around it.

I have only added support for landscape as I don't think it's reasonable to use this on portrait monitor. But if you feel like that's necessary then I can add that as well.


rxhanson:
Thanks! I changed it to Eighths instead of Eights just to be consistent with the rest of the window actions.

PR #763, opened 2022-02-21
Update russian language translation

dirondin:
Translate Todo section to Russian language and minor fixes with translation "Shortcut" word.

Also I can make translation to Belarusian language 🤍❤️🤍 (🇧🇾)


rxhanson:
Sure, feel free to add Belarusian. If you do, note that there are a lot of menu items that don't need to be translated since they aren't actually in the UI. Also, if you want license keys for Rectangle Pro & Multitouch, send me an email (as per the contribution policy).

PR #798, opened 2022-04-23
Add option to apply gaps to the main screen only

g4rb4g3:
Hi there 👋 
This PR adds another hidden preference to apply the screen gaps only to the main/primary screen.
I find it quite useful since I have uBar as dock replacement on the primary screen only. Without that the set gaps will be applied on non primary screens too which don't have any dock leaving some unused screen space.


rxhanson:
I made some updates on your fork/branch and thought that it would be automatically reflected in this PR, but it appears that it's not.

The changes are just simplifications that should have the same outcome, but I haven't tested them out. Let me know if you want me to generate a new PR if those changes seem good to you, or I could just merge this and make the changes myself.

Note that I got rid of your CG calls, as I think the NSScreen.screens.first check should be equivalent.
https://developer.apple.com/documentation/appkit/nsscreen/1388393-screens

g4rb4g3:
Thank you for the update. I'm still new to the whole Apple stuff. 😬 
Feel free to merge everything, I can give it a try afterwards.

rxhanson:
I merged it in and added my changes - feel free to pull the latest and try it out.

PR #799, opened 2022-04-23
Added support for corner sixth snap areas

davystrong:
Added support for snapping to corner sixths, as I like sixths for a wide screen but don't like keyboard shortcuts. I used similar logic to the two thirds snap areas, but starting from the corner and dragging inwards, to the thirds position, and the equivalent at the top.
I couldn't work out what `topLeftShort`, etc. meant and my logic works with and without the dock. Let me know if I need to add some logic to deal with "short" options.
I didn't add any region for centre sixths, as I can't think of any good gestures.
I also didn't add the logic for portrait mode. Would this be useful? I don't think I have any way to actually test this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Drag a window to the edge of the screen. When the mouse cursor reaches the edge 
 | Left or right edge, just above or below a corner       | Top or bottom half                     |
 | Bottom left, center, or right third                    | Respective third                       |
 | Bottom left or right third, then drag to bottom center | First or last two thirds, respectively |
+| Corners, then drag to respective third                 | Respective sixth                       |
 
 ### Ignore an app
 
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks for contributing. This looks good, but at first cut it should all go behind a feature flag, configurable behind a terminal command. To do this, you can just create another entry in the Defaults class, and you might as well make it an OptionalBoolDefault which allows it to be easily enabled in the code in a future release if desired. Then in the code you would just check if the default is userEnabled. If this is confusing just let me know and I can clarify or just merge your pr and add it. I think this is a good improvement, but don't want to spring it on users suddenly.

`topLeftShort` and corresponding snap areas are the top half / bottom half areas on the left & right edges near the corners. If you drag the window up the left side of the screen then you'll see it change to top half before top left corner.

That's fine to go without portrait mode equivalents at this time, especially since it's initially rolled out via feature flag.

davystrong:
Hi @rxhanson, thanks for that explanation, I've added it as you suggested.

rxhanson:
Thanks! I'm planning on rolling out a release pretty soon.

PR #807, opened 2022-04-30
fix: screenEdgeGapsOnMainScreenOnly logic inverted

g4rb4g3:
Hello again 😃 
It is about my [last PR](https://github.com/rxhanson/Rectangle/pull/798), you did some refactoring and inverted the logic by accident. 😬  Instead of applying the gaps to the main screen only it does it only for non main screens now. I fixed it with this PR.


rxhanson:
Thanks!

PR #812, opened 2022-05-08
Configure GitHub Actions to upload build as an artifact

Aaron-Rumpler:


.github/workflows/build.yml
```diff
@@ -1,18 +1,48 @@
...
+        plutil -convert xml1 - -o "$EXPORT_OPTIONS_PLIST" << EOF
+          {
+            "destination": "export",
+            "method": "mac-application"
+          }
+        EOF
+        
+        xcodebuild -exportArchive -archivePath "$BUILD_DIR/$XCODE_ARCHIVE" -exportPath "$BUILD_DIR" -exportOptionsPlist "$EXPORT_OPTIONS_PLIST"
+    - name: Resign App
+      run: codesign --force --deep -s "$CODE_SIGN_IDENTITY" "$BUILD_DIR/$APP_NAME"
```
Aaron-Rumpler, line 41: `RectangleLauncher` has issues with how `xcodebuild` signs the app. This works as a quick fix for that.

rxhanson:
Thanks! Excuse my naiveté, but how does this work? Without my credentials uploaded, who is it signed as? And since it's not notarized, gatekeeper would have to be bypassed for an install. If you would rather point me to docs, that's alright, too.

Aaron-Rumpler:
Setting `CODE_SIGN_IDENTITY="-"` causes Xcode to self-sign the compiled app (so not a Developer ID certificate, but does meet the code signing requirement for Apple Silicon).

Aaron-Rumpler:
I personally think a self-signed build is acceptable for development builds, but it is possible to set it up to sign with your Developer ID if that's desired (and potentially automate the release builds as well). `xcodebuild -help` has some documentation regarding that, and there's a documentation page [here](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow) as well.

.github/workflows/build.yml
Aaron-Rumpler: This is to do with the self-signing I believe, so might fix itself if you sign with your Developer ID.

.github/workflows/build.yml
```diff
@@ -1,18 +1,48 @@
...
+    - name: Make DMG
+      run: hdiutil create -srcdir "$BUILD_DIR" -volname "$DMG_NAME" "$DMG_FILE_NAME"
+    - name: Upload
+      uses: actions/upload-artifact@v2
+      with:
+        name: Build
+        path: $DMG_FILE_NAME
```
filterpaper, line 48:
```suggestion
        path: ${{ env.DMG_FILE_NAME }}
```
This should fix the [no file found error.](https://github.com/rxhanson/Rectangle/runs/6343413229?check_suite_focus=true#step:8:1)

filterpaper (changes requested):

.github/workflows/build.yml
Aaron-Rumpler: Oops, that would be from when I changed from `"${{ env.VAR_NAME }}"` to `"$VAR_NAME"`. Fixed.

PR #830, opened 2022-06-10
Some adjustments to the German translation

fabifabulousity:
The German descriptions have been adapted to the exact wording of the translations of system apps in macOS to ensure better consistency.
Also, prompts have been adapted to the "Du" form of address commonly used in macOS.


rxhanson:
Thanks!

Issue #829, opened 2022-06-08
macOS 13 Stage Manager option


decodism:
macOS 13 is only in early beta and there could be some changes but I think it would be nice to have an option to support Stage Manager. It should be enough to ignore the recent apps area on the left (about 375px on my screen), in a similar way to the todo mode on the right. It should be checked that this area is enabled (`defaults read com.apple.WindowManager GloballyEnabled` is `true` and `defaults read com.apple.WindowManager AutoHide` is `false`).

rxhanson:
Thanks! I do typically wait a little while before rolling in any changes, but if anyone else wants to make any changes with the proper `#available` macOS versioning then I'm good with rolling something in earlier and pushing out a new release.

Aaron-Rumpler:
From what I've seen online, the Stage manager sidebar also auto-hides if it's covered by a window, so it should probably be an option as to whether or not to ignore that sidebar. Though in my personal opinion, not covering that sidebar should probably be the default.

decodism:
I have looked a bit into this but there is an annoying issue: when restoring a scene, the windows moved programmatically with Rectangle are first restored to their last "manual" position before adopting their correct position.

Also, I'm not sure how wide the area should be reserved. On my screen:

- It needs to be at least about 47pt, otherwise the area hides itself
- There is a "magnetic edge" at about 129pt
- The area never seems to exceed about 191pt (including a 16pt padding)

Maybe the user should be given the choice. It could also be calculated dynamically by looking for the widest "window" of WindowManager.

hankhank10:
I have the beta installed on my secondary machine. It's very stable and in fact the only reason I wouldn't also install it on my primary machine would be that I rely on rectangle so much!

decodism:
I sent a feedback to Apple regarding the issue I encountered. I hope they take them into account.

dklad:
Not sure if it's because of a change in the latest version of Rectangle or the MacOS beta, but this issue appears to be resolved.

PR #834, opened 2022-06-23
Basic Stage Manager support

decodism:
As the issue described in #829 seems to have been solved, I propose a basic support of Stage Manager, that is the configuration of the size of the area reserved for recent apps.


rxhanson:
Thanks! This looks good to me, and I've merged it, although I have yet to install the beta and try things out. I'm going to wait until I do that, and work through a couple other issues before pushing out a release. The additional UI section also needs to be inserted into all the localization files, but I'm going to play around with Stage Manager so I can see if there's any way to get around the need for UI configuration.

PR #838, opened 2022-06-29
Add Japanese localization about state manager

Fus1onDev:
The wordings are adapted as much as possible to those of the official Apple website in Japan.


rxhanson:
Thanks!

PR #849, opened 2022-07-07
Updated German Strings

Wandang:
Not sure what Frontend and therefore context this is, so I updated it to the best of my abilities. (I use Rectangle on Mac and never saw a Launcher)


rxhanson:
Thanks!

PR #862, opened 2022-07-23
add quarter column snap windows

cedi:
Adding seemless support for snapping to thrids and quarterWidth snap areas.

This might sound a bit stupid at first, but here I am:
I recently got a super ultra-wide screen (49" 32:9) as my main screen and I use my "old" ultra-wide screen (34" 21:9) as a second monitor stacked on top of the super ultra-wide one.
The problem I'm facing is, while the thirds snap areas work extremely well on the ultra-wide screen, they are just too big for my super ultra-wide screen. I would prefer to have quarters instead.
I looked around with other software, but turns out: no other software supports snap-areas for quarter columns. Magnet does thirds and sixths columns (but I only noticed this after purchasing it).

Since Rectangle is OpenSource I decided to just try and implement the quarter snap areas myself and see if I can make it work. Turns out: I somehow made it work 🤣 
I hid  the quarter snap areas behind a feature-flag that needs manual enabling, since I suspect most people will use standard wide screens (16:9) or at best ultra-wide screens and therefore won't really need or want this feature.

Let me know what you think about it and please let me know if the code(-style/-quality) is okay. 
Actually, I never wrote a single line of Swift in my life before. But this bothered me so much, that I decided to just give it a shot.

Cheers,
Cedric

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -68,6 +68,7 @@ class Defaults {
     static let shortEdgeSnapAreaSize = FloatDefault(key: "shortEdgeSnapAreaSize", defaultValue: 145)
     static let cascadeAllDeltaSize = FloatDefault(key: "cascadeAllDeltaSize", defaultValue: 30)
     static let sixthsSnapArea = OptionalBoolDefault(key: "sixthsSnapArea")
+    static let quarterColumnsSnapAera = OptionalBoolDefault(key: "quarterColumnsSnapAera")
     static let eightsSnapArea = OptionalBoolDefault(key: "eightsSnapArea")
     static let stageSize = FloatDefault(key: "stageSize", defaultValue: 190)
 
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -68,6 +68,7 @@ class Defaults {
     static let shortEdgeSnapAreaSize = FloatDefault(key: "shortEdgeSnapAreaSize", defaultValue: 145)
     static let cascadeAllDeltaSize = FloatDefault(key: "cascadeAllDeltaSize", defaultValue: 30)
     static let sixthsSnapArea = OptionalBoolDefault(key: "sixthsSnapArea")
+    static let eightsSnapArea = OptionalBoolDefault(key: "eightsSnapArea")
     static let stageSize = FloatDefault(key: "stageSize", defaultValue: 190)
 
     static var array: [Default] = [
--- a/Rectangle/Snapping/SnappingManager.swift
+++ b/Rectangle/Snapping/SnappingManager.swift
@@ -408,19 +408,21 @@ class SnappingManager {
             let quarterWidth = floor(frame.width / 4)
             
             // check first quarter column
-            if loc.x >= frame.minX && loc.x <= frame.minX + quarterWidth {
-                if Defaults.eightsSnapArea.userEnabled {
-                    if let priorAction = priorSnapArea?.action {
-                        let action: WindowAction
-                        switch priorAction {
-                        case .bottomLeft, .bottomLeftSixth, .bottomLeftEighth:
-                            action = .bottomLeftEighth
-                        default: action = .firstFourth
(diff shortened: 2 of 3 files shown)


rxhanson:
Thanks for contributing!

I will dig into this in a little while, when I have a bit more time. For a long time, I've wanted to revisit this part of the `Snapping Manager` to see if I can improve it (as it's always been a bit of a pain to read through), and I'm starting to think I should do it sooner rather than later.

Rectangle/Defaults.swift
```diff
@@ -68,6 +68,8 @@ class Defaults {
     static let shortEdgeSnapAreaSize = FloatDefault(key: "shortEdgeSnapAreaSize", defaultValue: 145)
     static let cascadeAllDeltaSize = FloatDefault(key: "cascadeAllDeltaSize", defaultValue: 30)
     static let sixthsSnapArea = OptionalBoolDefault(key: "sixthsSnapArea")
+    static let quarterColumnsSnapAera = OptionalBoolDefault(key: "quarterColumnsSnapAera")
```
rxhanson, line 71: Typo, should be `quarterColumnsSnapArea`

Rectangle/Defaults.swift
```diff
@@ -68,6 +68,8 @@ class Defaults {
...
+    static let eightsSnapArea = OptionalBoolDefault(key: "eightsSnapArea")
```
rxhanson, line 72: Keeping with convention, should be `eighths`. It looks like only bottom eighths was implemented, in which case a more applicable name would be `bottomEighthsSnapArea`? Since the sixthsSnapArea also handles the top, it feels a little incomplete to only have the bottom eighths implemented.

Rectangle/Snapping/SnappingManager.swift
```diff
@@ -405,6 +405,27 @@ class SnappingManager {
         
         if loc.y >= frame.minY && loc.y < frame.minY + marginBottom && !ignoredSnapAreas.contains(.bottom) {
             let thirdWidth = floor(frame.width / 3)
+            let quarterWidth = floor(frame.width / 4)
+            
+            // check first quarter column
+            if Defaults.quarterColumnsSnapAera.userEnabled {
+                if loc.x >= frame.minX && loc.x <= frame.minX + quarterWidth {
+                    if Defaults.eightsSnapArea.userEnabled {
```
rxhanson, line 413: Quarter columns would need to be enabled in order for eighths to be applied. It seems like it would be more desirable to have these be configurable separate from each other?

Rectangle/Snapping/SnappingManager.swift
```diff
@@ -419,6 +440,32 @@ class SnappingManager {
...
+                    if let priorAction = priorSnapArea?.action {
+                        let action: WindowAction
+                        switch priorAction {
+                        case .firstFourth, .leftHalf:
+                            action = .firstThreeFourths
+                        case .lastFourth, .rightHalf:
+                            action = .lastThreeFourths
+                        case .bottomLeftEighth, .bottomRightEighth, .bottomCenterLeftEighth, .bottomCenterRightEighth:
+                            if Defaults.eightsSnapArea.userEnabled {
+                                action = .bottomCenterLeftEighth
```
rxhanson, line 456: I think there might be a logic error somewhere around this? I cannot get `bottomCenterLeftEighth` (or `bottomCenterRightEighth`) to trigger.

rxhanson (changes requested):
Sorry that this has taken so long for me to take a look at. As mentioned, I did want to clean the code in this class and pull things into functions to make things more readable, but I think it makes more sense to get things merged in first and worry about cleaning it up after.

rxhanson:
I actually went ahead and refactored all of the snapping manager logic here so that I could make snap areas configurable in the UI. I'll be merging it in soon and creating a release. I'll go ahead and add in the quarter columns as an option for the bottom snap area, but I think I'll leave out the eighths at first cut. Gonna close this one out, since the ground shifted out from under it. Thanks for contributing, though!

PR #875, opened 2022-08-04
Stage Manager: only one window case

### LINKED ISSUE #873: Ignore stage manager spage on secondary displays

macOS version: 13.0
Rectangle version: 0.57 (63)

```
2022-08-03T17:08:41+07:00: AX sizing proposed: (850.0, 1880.0), result: (804.0, 1354.0)
2022-08-03T17:08:41+07:00: AX position proposed: (2770.0, 180.0), result: (2770.0, 180.0)
2022-08-03T17:08:41+07:00: AX sizing proposed: (850.0, 1880.0), result: (850.0, 1880.0)
2022-08-03T17:08:41+07:00: maximize | display: (2750.0, -640.0, 890.0, 1920.0), calculatedRect: (2770.0, 180.0, 850.0, 1880.0), resultRect: (2770.0, 180.0, 850.0, 1880.0), srcScreen: , destScreen: , resultScreen: 
```

Stage manager recent apps only shows in the primary display, so resizing windows on the other displays should not count the gap.

decodism:
Fixes #873

### DIFF
--- a/Rectangle/Utilities/StageUtil.swift
+++ b/Rectangle/Utilities/StageUtil.swift
@@ -14,5 +14,23 @@ class StageUtil {
         
         return stageDefaults.bool(forKey: "GloballyEnabled") && !stageDefaults.bool(forKey: "AutoHide")
     }
+
+    static func stageWindowPresent() -> Bool {
+        let options = CGWindowListOption(arrayLiteral: .excludeDesktopElements, .optionOnScreenOnly)
+        if let windowInfo = CGWindowListCopyWindowInfo(options, 0) as? Array<Dictionary<String,Any>> {
+            for infoDict in windowInfo {
+                if let bounds = infoDict[kCGWindowBounds as String] as? [String: CGFloat] {
+                    guard let name = infoDict[kCGWindowOwnerName as String] as? String,
+                          let w = bounds["Width"],
+                          let h = bounds["Height"]
+                    else { continue }
+                    if name == "WindowManager" && w == 64 && h == 64 {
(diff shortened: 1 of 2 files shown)


Rectangle/Utilities/StageUtil.swift
```diff
@@ -14,5 +14,23 @@ class StageUtil {
         
         return stageDefaults.bool(forKey: "GloballyEnabled") && !stageDefaults.bool(forKey: "AutoHide")
     }
+
+    static func stageActuallyVisible() -> Bool {
```
rxhanson, line 18: Thanks for consistently keeping on top of the feature you introduced! I really appreciate it. A couple notes on this one:

- stageActuallyVisible is a slightly confusing name for the uninitiated, but I'm not picky and it's fine if you want to leave it. Maybe something like stageWindowPresent
- CGWindowListCopyWindowInfo can cause delay for some users on certain macs. Calls to this function typically should be used sparingly. Checking the name/w/h is also potentially fragile. If there's another way we can do this, it might be worth exploring that first. At the very least, we would probably want to only execute this if the defaults for stage manager show it enabled.

rxhanson: Apologies for the delay, thanks for the update. You're correct that Rectangle doesn't cache windows. I'm going to go ahead and roll this in. It's probably not too likely that Macs running Ventura will be impacted by one additional call to CGWindowListCopyWindowInfo here, so I'm willing to take the gamble since I can easily troubleshoot back here if needed.

PR #881, opened 2022-08-16
Stage Manager: avoid footprint window being moved

decodism:
When Stage Manager is enabled, it seems that new windows are moved away from the lateral edges of the screen, which is an issue for the footprint window. The proposed fix is to "show" the window before adjusting its frame.


rxhanson:
Thanks!

PR #895, opened 2022-08-26
Stage Manager: snap windows from recent apps

### LINKED ISSUE #853: Miniature windows from Stage Manager cannot be snapped

macOS version: 13.0 Beta
Rectangle version: 0.57
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): -

Stage Manager shows any windows that are not currently on screen in their miniature form on the left.

To get them onto your current "stage" you can drag these miniatures onto the screen to where you want.

However you cannot directly snap these miniature windows when dragging from stage manager, you have to drag them onto the stage first, then drop them, to then snap them to an edge.

It would be nice if you could directly drag the miniatures from Stage Manager to the right edge for example to make them occupy the right half of the stage.

decodism:
Fixes #853

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -62,6 +62,14 @@ class AccessibilityElement {
                     }
                     return windowElement
                 }
+                if windowElement.getIdentifier() == windowInfo.id {
+                    if Logger.logging {
+                        let app = NSRunningApplication(processIdentifier: windowInfo.pid)?.localizedName ?? ""
+                        Logger.log("Window under cursor fallback matched: \(app) \(windowInfo)")
+                    }
+                    // In case the window is in Stage Manager recent apps
+                    return FallbackAccessibilityElement(windowElement.underlyingElement)
+                }
             }
         }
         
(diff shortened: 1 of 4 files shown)


Rectangle/AccessibilityElement.swift
```diff
@@ -70,6 +81,43 @@ class AccessibilityElement {
...
+                        let boundsRect = NSMakeRect(x, y, w, h)
+                        windowInfos.append(CGWindowInfo(id: id, pid: pid, rect: boundsRect))
+                    }
+                }
+            }
+            if var resultWindowInfo = windowInfos.first(where: { $0.rect.contains(location) }) {
+                if StageUtil.stageCapable() && StageUtil.stageEnabled() && StageUtil.stagePresent(windowInfo) {
+                    // In case the window is in Stage Manager recent apps
+                    var prevWindowInfo: CGWindowInfo?
+                    while prevWindowInfo?.id != resultWindowInfo.id {
```
rxhanson, line 107: I feel like I'm missing something here with this while loop. It appears to me that it is unnecessary, but perhaps you can elaborate on what it's doing?

rxhanson (commented):
Looks good! Just one comment to talk through.

Rectangle/AccessibilityElement.swift
decodism: You can only drag the foreground window of a scene from the recent apps. This loop is used to find this window.

rxhanson: Hmm. Sorry to keep pushing on this, but when I test this out, I never get a different value for the resultWindowInfo (there's no iteration in the while loop because it gets the same value twice and exits, and it's not straightforward to me how it would obtain different values in any different scenario). Even if I remove this block entirely, I can still drag & snap the windows from recent apps because it appears to get the correct windowId. Perhaps there is an edge case that I'm not hitting that would justify having this?

The reason why this caught my attention in the first place is because 1) while loops that don't have an intuitive exit are too easy to end up in an infinite loop (if not now, maybe with other additional refactoring later), and 2) I pretty much never force unwrap optionals. Both of these scenarios are potential for application crashes, and even if your code here was straightforward to me, I would have still suggested removing both of those. In other apps I've written, places where I was certain there was no chance of a force unwrap causing a crash, still caused a crash due to some very tricky memory management from macOS that had released memory I thought wouldn't get released. One could argue that a crash is desirable in certain scenarios, but I rarely have those scenarios - I'd almost always want the app to still chug along even if there's something like this that isn't quite right.

rxhanson: Thanks for the clarification, I understand now. As for the while loop, I'm good with it as-is. I appreciate that you've continued the work on Stage Manager!

rxhanson:
And here's the analyzed crash report:
[AnalyzedCrashReport.txt](https://github.com/rxhanson/Rectangle/files/9976929/AnalyzedCrashReport.txt)

Issue #902, opened 2022-08-30
Dragging window to top sometimes triggers Mission Control


AndrewHaglund:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 12.5.1 (21G83)
Rectangle version: v0.58 (64)

Preferences:
- Snap windows by dragging is enabled
- Top is set to Maximize

When I drag a window to the top of the screen it often triggers Mission Control.

Seems like if I slowly drag up and my cursor hits the top of the screen and I stop — I will see the "snap windows" overlay and everything is good. However, if I'm doing this quickly my mouse goes up "too far" then Mission Control is triggered. I can "slam" the cursor to the other edges and Rectangle works as expected, but not the top gesture.

As a side note: I am a previous user of BetterSnapTool and I did not see this issue so I wonder if somehow they override macOS' drag-to-top gesture. (If I could disable Mission Control entirely, I would, but I don't know if there is a way to do that.)

Attached is a quick video of this issue. It happens to me 99% of the time.

https://user-images.githubusercontent.com/5075921/187516606-21d5aec7-4ec1-4f2b-8c99-f4b05c084bcb.mov

rxhanson:
This comes up from time to time. I don't quite have the bandwidth to add it in at the moment, but welcome a good pull request for it. I haven't tried out what BST does, but my guess is that it detects the velocity of the drag and freezes the cursor at the top of the display unless the velocity is high enough.

Shock9616:
I am also having this issue, although I would like to add that it works as expected if the Rectangle preferences window is open. As soon as I close it though Mission Control keeps activating when I try to maximize an app.

rxhanson:
@Shock9616 did you check the box in the snap areas tab in the prefs window to disable fast dragging to Mission Control?

Shock9616:
Yes I did. Here's a screen recording, except now it just doesn't seem to work at all. I must not have been doing it the same way when I commented previously. One thing that might be the issue is that I am on the MacOS Ventura public beta, but I'm pretty sure this was an issue for me before I updated as well.

https://user-images.githubusercontent.com/28808006/193656552-5654135b-7479-42c2-9334-16e3d0021bde.mov

rxhanson:
Closing out old issues. In a previous macOS release, it appeared to me that Apple adjusted this so Mission Control wasn't so easy to trigger so I never made any changes for this.

PR #904, opened 2022-09-02
Disable fast dragging to Mission Control

decodism:
Related to #902. Maybe change the sensitivity which is the `-25` value?


rxhanson:
Nice! I'll review this one and your other pull request this weekend, and am planning on making a new release in the next week or so.

rxhanson:
Awesome, this looks good to me. There's only one adjustment that I'd like here, and it's that I'd like to keep the event monitor intact and be able to switch between the two implementations depending on this setting. The reason is that if an event tap gets disabled and restarted (due to the timeout case that you have properly handled here 👍), certain Macs can have an input (like a click) discarded. In my experience, event taps can be disrupted by other apps. If we only use the event tap when the user desires this functionality, it both limits the potential for a discarded input that could be caused by another app, and it makes debugging something in this category much easier to narrow down.

With that said, I'm ok with merging this in and implementing the remainder of work here myself, or I can leave it up to you - whichever you prefer.

decodism:
It should be good.

rxhanson:
Thanks! I appreciate the follow on effort to keep the PassiveEventMonitor in place. I might readjust the checkboxes so that they are all in line.

PR #925, opened 2022-09-15
Execute action by URL

decodism:
For basic scripting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Drag a window to the edge of the screen. When the mouse cursor reaches the edge 
 1. Focus the app that you want to ignore (make a window from that app frontmost).
 1. Open the Rectangle menu and select "Ignore app"
 
-## Execute an action by URL
+## Execute an action by URL (Arriving in v0.60)
 
 Open the URL `rectangle://execute-action?name=[name]`. Do not activate Rectangle.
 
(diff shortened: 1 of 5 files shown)


rxhanson:
Nice! I didn't think about that way to do it. I'll start sponsoring you through GitHub sponsors if you wanted to set it up.

sentiasa:
Can you please show an example for this?

When I tried `open -g "rectangle://execute-action?name=left-half"`

```
No application knows how to open URL rectangle://execute-action?name=left-half (Error Domain=NSOSStatusErrorDomain Code=-10814 "kLSApplicationNotFoundErr: E.g. no application claims the file" UserInfo={_LSLine=1489, _LSFunction=runEvaluator}).
```

rxhanson:
@sentiasa this likely means that macOS hasn't made the association with the app. Maybe try uninstalling the app and reinstalling, or restarting your Mac.

sentiasa:
@rxhanson great, it worked. I run it in my terminal and it placed the terminal successfully. 

I was looking to achieve, for example, open phpstorm and place it left-half. Would this approach work with this? 
If so, how can I give application name to this command?

Maybe something like:
`open -g "rectangle://execute-action?name=left-half&app=phpstorm"`

Or is it not possible to achieve this with execute action command?

rxhanson:
@sentiasa These commands in Rectangle are the equivalent of executing the keyboard shortcut; there is no support for associating an app with the command at this time.

You can achieve this in [Rectangle Pro](https://rectangleapp.com/pro) (paid), by creating an "App Layout" and triggering that App Layout via URL (see https://github.com/rxhanson/RectanglePro-Community#programmatically-triggering-window-sizes--positions)

For future conversation on this topic, please create a new discussion in GitHub.

PR #952, opened 2022-10-17
Stage Manager: fix multi-screen

### LINKED ISSUE #928: Stage Manager with multiple monitors
### Discussed in https://github.com/rxhanson/Rectangle/discussions/927

<div type='discussions-op-text'>

<sup>Originally posted by **rusaito** September 25, 2022</sup>
I have a three-monitor set-up, and the stage manager (ventura beta) recent app area (I set to 175px) is correctly on the left side for my (1st) middle, and (2nd) left monitor. My (3rd) right monitor has the stage manager recent app area buffer on the right side instead of the left. In other words the windows default to cover up the stage manager and leave the same 175px buffer on the opposite (right side) </div>

decodism:
Tries to fix #928.


rxhanson:
Thanks! I'll probably roll out another release next week depending on anything else I'm looking to add in.

Issue #950, opened 2022-10-15
"Disable fast dragging to Mission Control" alters clicking behavior in Illustrator


kerryj89:
macOS version: 12.6
Rectangle version: v0.60 (66)

On latest Adobe Illustrator 2022 (I don't know if this is a new issue as I don't use it often), the "Disable fast dragging to Mission Control" option alters a regular click action (which selects the element) into a delayed click -> drag -> drop (wherever the cursor is after about a second).

What's strange is that in the video capture it seems like I pause after clicking (while "Disable fast dragging to Mission Control" option is enabled) while in reality I'm moving my cursor around randomly just like I do with the option disabled.

https://user-images.githubusercontent.com/1977804/196008304-339c011c-3615-4dc2-b493-1349a5b55877.mp4

rxhanson:
Thanks for reporting. The implementation of this feature is susceptible to this type of bug. It’s hard to tell where this bug might appear, so it’s helpful to know that you see it with Illustrator. 

It’s likely that there is no way to fix the bug with the current implementation, and the best solution might just be to notify Illustrator users and provide a way to disable it when Illustrator is frontmost. There also might be potential for an alternative implementation. 

It also makes sense that you would see some cursor lag immediately upon checking the box because this feature enables an alternate way to obtain mouse events, and registering this event listener can cause that.

kerryj89:
Thank you!  I agree it might be nice to have some kind of warning or notice in the future as it took a while to hunt down why this was happening.  I'll close the issue and maybe others who have the same or similar issue will see the solution here is to disable the option for now.

patrickrushton:
Thanks for reporting this one. I was having an issue with Logic Pro X when adjusting the cycle length or using the Marquee tool would behave erratically as if there was a delay before responding to a mouse up event. Weirdly it was only an issue if not running Logic in full screen. 

I had put it down to my Wacom driver. But it turns out that turning off the "Disable fast dragging to Mission Control" resolves the issue in Logic.

rxhanson:
I pulled this setting from the UI in v0.61 since the issue is hard to track down as a user, and I don't know what apps might be affected by it. A pull request has been merged that might fix it, but I don't plan on pushing out a new release for a little while. If you're up for testing it on an autogenerated, un-notarized build (you'd need to bypass gatekeeper for it), then you can try it here: https://github.com/rxhanson/Rectangle/actions/runs/3365970567

[...]

PR #955, opened 2022-10-19
Add a warning for "Disable fast dragging to Mission Control"

decodism:
Related to #950


rxhanson:
Thanks for following up on this. I've been thinking about it, and was considering making this a hidden, terminal command enabled feature just to reduce the number of areas where users can potentially shoot themselves in the foot. Having it as a terminal command typically forces users to read through potential side effects. I have had a lot of users skip reading the subtext on checkboxes, like the one for relaunching the app when the menu bar icon is hidden. What are your thoughts on it?

decodism:
I'm not against it.

rxhanson:
Thanks. I decided to pull it out in v0.61, but I would like to have it back in the UI at some point. If it were just Adobe Illustrator that had the issues, I was thinking about applying some messaging specific to Illustrator (if installed on the users machine), but I had multiple users email me directly saying that it was affecting text selection in multiple apps. I'll dig into this a little deeper in an upcoming release and see if I can get something narrowed down.

Issue #958, opened 2022-10-25
 Catalan translation is incomplete despite the translation being complete in the code 


fv3rdugo:
I found a similar issue documented #254 

macOS version: 12.6 (21G115)
Rectangle version: 0.60
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): Non applicable

Some strings in the the catalan version aren't translated, eg: "Left Half", "Center Half", "Top Left", "Almost Maximize", ...
I've checked the localization files, and all the strings seem to be correctily translated in ca.lproj

My specific location is:   `osascript -e 'user locale of (get system info)'`
 ca_ES

rxhanson:
Hmm. I'll copy over ca into ca_ES, just like in the other issue (thanks for referencing it).

rxhanson:
Thanks for looking to do the PR. Closing this one out, but feel free to reopen if you hit any snags.

fv3rdugo:
I'm sorry, the problems have not been fixed with the new version (Versió 0.61 (67)). Could it be ca_ES instead of ca-ES, I'm not sure. Left Half, Top Half, Top Left, Last Third, Smaller, Move to the Edge, ...are some examples.
<img width="249" alt="Captura de pantalla 2022-10-29 a les 10 08 46" src="https://user-images.githubusercontent.com/3403589/198821158-1951b000-b5d2-44d9-88f3-48b81c033f81.png">
<img width="298" alt="Captura de pantalla 2022-10-29 a les 10 14 32" src="https://user-images.githubusercontent.com/3403589/198821271-b6e451eb-7898-49f3-a706-809e81b8ac45.png">

rxhanson:
Is this identical to what you had prior to v0.61? If so, then we can do away with the additional translation file and just fix the existing "ca" one. 

The issue is that there are characters that are breaking the quote terminations on the strings. I'm actually not sure which characters are the problematic ones, but it's probably a single quote, and using a \ doesn't seem to do anything to it? I could be wrong though. Do you have any idea?

fv3rdugo:
Yes, you're right, there is a missing end quote, only one. Sorry for the mistake. I think this was the original error, and not the ca_ES locale. Maybe removing it is a the best idea. Thank you for your quick reply.

PR #978, opened 2022-10-31
Missing end quote in catalan translation

fv3rdugo:
Yes, you're right, there is a missing end quote, only one. Sorry for the mistake. I think this was the original error in issue #958.


rxhanson:
Sorry for my lack of communication. I already made this change so this PR doesn't change the code. I'll remove the ca-ES as well. Thanks!

PR #979, opened 2022-11-02
Small change to normalise the top-right string with the other quarts

climent:
Really small change, but it was bothering me that the other quarters were using a different approach to the string.


rxhanson:
Thanks!

Issue #981, opened 2022-11-02
Maximum "Stage Manager recent apps area" value is too small


hsjunn:
- [X] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: Ventura 13.0
Rectangle version: 0.61 (67)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): N/A

In my iMac Pro 2017 with the "default" display resolution settings, Stage Manager uses 214 pixels for its recent apps area,
which is larger than the current maximum value allowed by Rectangles i.e., 200.
Could you please increase the maximum value?

My current (working) workaround is:
- export the preferences to json
- manually edit the "stageSize" float value
- save
- import the edited json

edit: I understand Rectangles doesn't need to exactly replicate that value, but I feel sorta compulsive about that.

rxhanson:
Agreed. I'll slate this for the next release.

YouRik:
I think an tick box option to always keep the space for stage manager would be nice, not only when there are recent apps in the stage manager area.
I'm suggesting this because when you have a window A maximized, then open a new window B, window A gets moved into the miniature view. When you then switch back to A, the stage manager area is hidden because A has not resized to account for the stage manager area and B gets completely hidden, so it's necessary to resize A manually.

rxhanson:
@YouRik The way to achieve what you're after is just to set a screen edge gap, using a terminal command (and restart the app):
https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md#setting-gaps-at-the-screen-edges

Then set stage manager recent apps area to 0.

Closing this one out since it was merged a while back.

YouRik:
> @YouRik The way to achieve what you're after is just to set a screen edge gap, using a terminal command (and restart the app): https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md#setting-gaps-at-the-screen-edges
> 
> Then set stage manager recent apps area to 0.
> 
> Closing this one out since it was merged a while back.

Thank you, I was not aware of the advanced settings through the terminal.

Issue #854, opened 2022-07-14
Snapping window to left half does not work with Stage Manager


leodr:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 13.0 beta
Rectangle version: 0.57
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): -

Dragging a window to the left edge of the screen puts it into Stage Manager as a minimized window.

AFAIK it is currently not possible on Ventura to snap a window to the left half of the screen without using the keyboard shortcuts.

rxhanson:
If you drag it to an area above or below the Stage Manager area then it's possible. I don't think there's a way to prevent it from putting it into the Stage Manager in its area, so I think the proper path forward would be for Rectangle to not display the footprint if the window is dragged into the Stage Manager area. Unfortunately, I don't think there is a way to get the exact size/location of the Stage Manager.

rxhanson:
(Housekeeping closing)

PR #999, opened 2022-11-15
Stage Manager

decodism:
- I can't reproduce #989. It looks like this is due to an infinite loop that occurs when the window ID cannot be determined directly. This ID is now cached.
- There is now a detection of the windows belonging to the strip which allows to disable their snapping when needed as suggested in #854.
- The maximum width of the recent apps area has been increased for #981.


decodism:
Besides, maybe the missionControlDragging and dragFromStage defaults could be enabled.

rxhanson:
Thanks! The refactoring was a great improvement. I'll enable `dragFromStage`, but `missionControlDragging` still appears to affect at the very least InDesign. I'm going to draft another pre-release with this change in, as don't have a chance to thoroughly test it for the next couple weeks.

PR #1007, opened 2022-11-22
Stage Manager: clean up workaround

decodism:
It seems to be unnecessary for the main window since at least the release of macOS 13.


rxhanson:
Thanks!

PR #1008, opened 2022-11-25
Add footprintAnimationDurationMultiplier default

decodism:
Add optional animations to the footprint window.

### DIFF
--- a/Rectangle/Snapping/SnappingManager.swift
+++ b/Rectangle/Snapping/SnappingManager.swift
@@ -10,6 +10,7 @@ import Cocoa
 
 struct SnapArea: Equatable {
     let screen: NSScreen
+    let directional: Directional
     let action: WindowAction
 }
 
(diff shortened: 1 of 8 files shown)


rxhanson:
Nice! I like it.

rxhanson:
I might add in a checkbox for this in Snap Areas tab, where checking the box would set it to something like 0.75 & unchecking would set it to 0.

decodism:
I agree. I understand you want to keep the settings minimalist but this one might be worth it.

Aaron-Rumpler:
Might be cleaner to scale the multiplier so that the default when enabling it can be 1, and that gets scaled to whatever you decide the default speed should be.

Rectangle/Snapping/SnappingManager.swift
```diff
@@ -270,9 +271,21 @@ class SnappingManager {
                         if box == nil {
                             box = FootprintWindow()
                         }
-                        box?.setFrame(.zero, display: false)
-                        box?.makeKeyAndOrderFront(nil)
-                        box?.setFrame(newBoxRect, display: true)
+                        if Defaults.footprintAnimationDurationMultiplier.value > 0 {
+                            if !box!.isVisible, let origin = getFootprintAnimationOrigin(snapArea, newBoxRect) {
```
Aaron-Rumpler, line 275: The animation's getting caught up here if the `FootprintWindow` hasn't closed yet (which only happens 0.3 seconds after the `box.close()` call), as it's still `isVisible`, even though it's completely transparent. So if you move a window away from a snapping area and straight back onto it within 0.3 seconds, the animation starts from where the window was when the window was moved away.

decodism: Good catch. I have created a new PR.

PR #1032, opened 2022-12-15
Allow configuration option for stageSize to be treated as a percentage if less than 1

benminer:
Hi!

Love this app, thanks for all your work on this. 

Since macOS Ventura, I️ am a huge fan of Stage Manager, but I️ switch from my desktop monitor setup to my laptop quite often. With the screens being vastly different sizes, the absolute pixel width value that Rectangle currently allows for is too much when using my laptop dimensions. Like, setting it to the perfect width when on my monitor resulted in the stage area being too large on my laptop. 

This PR adds an option to set a percentage of the screen width rather than a simple absolute value. If the percentage is set (defaults to 0), it is used over any set "stageSize" value. I️ did not add UI for this, as it is quite niche, just a reference from the config file if set. 

Locally my build works fine, but I️ was _not_ able to fully test as I️ couldn't get the debug .app file to be authorized via Accessibility (probably due to me having the release Rectangle installed as well).

Let me know if there are any issues or questions! I️ would absolutely love if this got shipped, it's been driving me nuts!

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -69,7 +69,7 @@ class Defaults {
     static let cascadeAllDeltaSize = FloatDefault(key: "cascadeAllDeltaSize", defaultValue: 30)
     static let sixthsSnapArea = OptionalBoolDefault(key: "sixthsSnapArea")
     static let stageSize = FloatDefault(key: "stageSize", defaultValue: 190)
-    static let stagePercentageWidth = IntDefault(key: "stagePercentageWidth", defaultValue: 10)
+    static let stagePercentageWidth = IntDefault(key: "stagePercentageWidth", defaultValue: 0)
     static let dragFromStage = OptionalBoolDefault(key: "dragFromStage")
     static let landscapeSnapAreas = JSONDefault<[Directional:SnapAreaConfig]>(key: "landscapeSnapAreas")
     static let portraitSnapAreas = JSONDefault<[Directional:SnapAreaConfig]>(key: "portraitSnapAreas")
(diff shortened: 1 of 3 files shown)


rxhanson:
Thanks for contributing. 

Rather than adding another defaults variable, we could just check and see if stageSize is less than 1 and use that value as a percentage. With so may defaults variables, I prefer to not add one wherever possible.

benminer:
> Thanks for contributing.
> 
> Rather than adding another defaults variable, we could just check and see if stageSize is less than 1 and use that value as a percentage. With so may defaults variables, I prefer to not add one wherever possible.

Great idea, will do.

benminer:
Done!

rxhanson:
Looks good to me. 

Two minor things:
1. I will change the variable name `adjustValue` to `stageSize` if you don't.
2. I will add a section to https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md if you don't.

Not a big deal either way, just let me know if you're going to do it and I'll wait to merge the PR.

benminer:
> Looks good to me.
> 
> Two minor things:
> 
> 1. I will change the variable name `adjustValue` to `stageSize` if you don't.
> 2. I will add a section to https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md if you don't.
> 
> Not a big deal either way, just let me know if you're going to do it and I'll wait to merge the PR.

I️ can take care of it, I️'ll have some this evening.

benminer:
> > Looks good to me.
> > Two minor things:
> > 
> > 1. I will change the variable name `adjustValue` to `stageSize` if you don't.
> > 2. I will add a section to https://github.com/rxhanson/Rectangle/blob/master/TerminalCommands.md if you don't.
> > 
> > Not a big deal either way, just let me know if you're going to do it and I'll wait to merge the PR.
> 
> I️ can take care of it, I️'ll have some time this evening.

Done

rxhanson:
Thanks! It might be a few weeks before I make the next release, depending on how many changes get rolled in or if any important bugs come up.

Aaron-Rumpler:
> Locally my build works fine, but I️ was _not_ able to fully test as I️ couldn't get the debug .app file to be authorized via Accessibility (probably due to me having the release Rectangle installed as well).

Those permissions seem to be tied to the app signature (presumably the team id) as well as the bundle id. Easiest fix is to make sure to remove Rectangle from the accessibility section (or run `tccutil reset All com.knollsoft.Rectangle` from the terminal) between exiting one copy and opening another.

PR #1067, opened 2023-02-06
Todo Mode improvements

decodism:
- Correction of the target of the toggling selector when the settings have the first responder.
- Disable reflow when mode is disabled.
- Automatic reflow after changing settings.
- Ignore sidebar area for actions on the sidebar window.
- The sidebar window can be resnapped to the sidebar area.
- Add a shortcut to toggle mode.
- Save shortcuts to the config.
- Add a setting for the sidebar side.


rxhanson:
Looks good at a quick read through. I'll test it out soon.

One thing to note is that some users did use the reflow with todo mode disabled, so this is going to appear as broken to those users. I've always felt like that's a weird side effect that doesn't quite fit with what Todo mode is trying to achieve, and it is a really small percentage of users, so I'm good with your change.

rxhanson:
Thanks! I'm going to adjust the settings tab of the prefs window so it looks a little more tidy before releasing this, though.

PR #1083, opened 2023-02-24
Fix initial Todo shortcuts display

### LINKED ISSUE #1082: TODO Mode broken on Version 0.66 (72)
macOS version: 12.6.3
Rectangle version: Version 0.66 (72)

TODO Mode seems to move the window the the left side instead of the right side and other windows don't respect the TODO window boundaries anymore.

decodism:
- Old bug: Todo shortcuts display in settings remained empty as long as Todo Mode was never shown in menu.
- Fixes #1082

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -138,7 +138,7 @@ class AccessibilityElement {
         return childElements?.filter { $0.role == role }
     }
     
-    fileprivate var windowId: CGWindowID? {
+    var windowId: CGWindowID? {
         wrappedElement.getWindowId()
     }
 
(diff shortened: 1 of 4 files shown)


rxhanson:
Thanks for following up on that issue. 

On the initial initial shortcut display, there was intention behind having it that way, but I agree that it feels like a bug. 

By only initializing the shortcut when Todo mode was added to the menu, Rectangle wasn't adding a keyboard shortcut unless the user showed intent to use Todo mode. Todo mode wasn't in the initial feature set, and I didn't secretly bind a keyboard shortcut if the user did not show interest (and some users just don't go through the settings tab). 

Since a user would still have to enable todo mode in the menu in order to set the todo application, it's actually acceptable to me to leave it the way it was before this change. Since the shortcuts aren't cleared when todo mode is removed from the menu, it is straightforward that those shortcuts are still registered. With your change, the shortcuts are displayed but not registered until

rxhanson:
Thanks! This covers the edge case I was concerned with - the case where a user might be confused about a keyboard shortcut binding that either isn't active or is unintentionally active.

PR #1084, opened 2023-02-27
Todo Mode refinements

decodism:
- Be able to switch the Todo window explicitly rather than having to close the previous one.
- Ignore the Todo window if it doesn't exist in the calculation of the frames of the other windows.


rxhanson:
I like it. I'll probably roll out a release sometime this week with your changes.

PR #1095, opened 2023-03-03
feat: 🌐 Add 🇺🇦 Ukrainian translation

kopach:
Ref: https://github.com/rxhanson/Rectangle/discussions/1094

### DIFF
--- a/Rectangle/uk.lproj/Main.strings
+++ b/Rectangle/uk.lproj/Main.strings
@@ -1,550 +1,550 @@
 
 /* Class = "NSTextFieldCell"; title = "Last Two Thirds"; ObjectID = "08q-Ce-1QL"; */
-"08q-Ce-1QL.title" = "Last Two Thirds";
+"08q-Ce-1QL.title" = "Останні дві третини";
 
 /* Class = "NSTextFieldCell"; title = "Top Right"; ObjectID = "0Ak-33-SM7"; */
-"0Ak-33-SM7.title" = "Top Right";
+"0Ak-33-SM7.title" = "Верхній правий кут";
 
 /* Class = "NSTextFieldCell"; title = "0 px"; ObjectID = "0eh-6G-rMp"; */
-"0eh-6G-rMp.title" = "0 px";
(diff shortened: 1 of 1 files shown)


Rectangle/uk.lproj/Main.strings
```diff
@@ -1,550 +1,550 @@
...
+"Left half, top/bottom half near corners" = "Ліва половина, верхня/нижня половина близько кутів";
+"Right half, top/bottom half near corners" = "Права половина, верхня/нижня половина близько кутів";
+"Thirds, drag toward center for two thirds" = "Третини, перетягніть у центр для двох третин";
+"Thirds, top/bottom half near corners" = "Третини, верхня/нижня половина близько кутів";
+"Left or right half" = "Ліва або права половина";
+"Top sixths from corners; maximize" = "Верхні шості від кутів; максимізувати";
+"Bottom sixths from corners; thirds" = "Нижні шості від кутів; третини";
+"Fourths columns" = "Четвертини стовпців";
+
+"Go to System Settings → Privacy & Security → Accessibility" =
```
rxhanson, line 546: Everything else looks good, just this have this missing one.

rxhanson (changes requested):
Thanks! Just one missing one near the end.

Rectangle/uk.lproj/Main.strings
kopach: oh, sorry, missed that somehow. Fixed now

kopach:
> Thanks! Just one missing one near the end.

fixed @rxhanson. Please, take a look once again, thanks

Btw, how about this file https://github.com/rxhanson/Rectangle/blob/aa434500ec61292c5bf7da92c8fb0de477abdf54/RectangleLauncher/uk.lproj/Main.strings 
Should it also be translated?

rxhanson:
Thanks! That file doesn't need to be translated (it's the embedded helper app without a UI that is used to launch the app on startup for macOS < 13).

rxhanson (approved):

PR #1100, opened 2023-03-08
Add double-click title bar

decodism:
Implements #1097. The detection of the title bar is a bit worse than the one of the system.


rxhanson:
Thanks! I think what I prefer here is to replace the config in the UI with one checkbox that is for letting Rectangle manage the double click in the title bar to maximize/restore, and let users further configure it via Terminal command if they really want it to do other things. I didn't check but maybe there is an NSUserDefault that we can check to see if they have it enabled in System Settings, and prompt the user to disable it if it is enabled when a user checks this one checkbox in Rectangle.

PR #1101, opened 2023-03-08
Stage Manager: footprint window workaround

### LINKED ISSUE #1099: Footprint preview is drawn incorrectly when Stage Manager is enabled

macOS version: 13.2.1 (22D68
Rectangle version: v0.66 (72)

I have Stage Manager enabled, but the slider "Stage Manager recent apps area" set to 0px.

I would expect the footprint animations to use the entire screen area, however when I drag a window to a snap area, the footprint leaves a gap where the recent apps would appear - the width and height appear correct while the x, y origin is offset. 

After completing the drag, the window is resized as expected (i.e. it doesn't leave a gap for stage manager recent apps) - it's just the footprint that seems to be drawn incorrectly.

decodism:
Tries to fix #1099.

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -262,6 +262,7 @@ extension AccessibilityElement {
             return windowElement
         }
         if let info = getWindowInfo(position) {
+            // Workaround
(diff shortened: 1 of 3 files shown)


rxhanson:
Awesome, thanks!

PR #1105, opened 2023-03-10
Update Simplified Chinese localization

Jerry23011:
### Changes
- Add new translations
- Adjust spacing (according to conventions, there should be a space between English letters and Chinese characters)


rxhanson:
Thanks! Since macOS < 13 is still supported, the `Preferences` strings are still valid, and any instances in Ventura where `Preferences` is still displayed in the app, I need to fix it in the code rather than here in the localization file.

Jerry23011:
> Thanks! Since macOS < 13 is still supported, the `Preferences` strings are still valid, and any instances in Ventura where `Preferences` is still displayed in the app, I need to fix it in the code rather than here in the localization file.

Got it, I just altered the translation only, so I'll just revert that particular change and wait for the change in code :)

rxhanson:
Excellent, thanks!

PR #1121, opened 2023-03-18
Update Japanese translation

ryonakano:
### Changes Summary

- "Settings" should be translated to just "設定" instead of "環境設定"
  - L405 and L621
- Fix mistranslation
  - L603: `左中央` to `中央上`
  - L612: `左中央` to `中央下`
- Add translations for new strings
- Follow the Apple's translation against "window": use "ウインドウ" instead of "ウィンドウ"


rxhanson:
Thanks!

Issue #566, opened 2021-09-08
Export does not write a file


roblight:
- [X] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: macOS 11.5.2 (20G95)
Rectangle version: 0.48 (53)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): Empty logs

I tried various directories other than the default (~/Documents) to no avail.

Excellent app - thank you so much!

rxhanson:
Sorry for the delayed response on this one. I'm unable to reproduce this and since no one else has reported this, it is likely something specific to your machine. Is there anything unusual about your setup? I have noticed that sometimes the macOS save dialog has had strange behavior in the past, maybe just uninstalling and reinstalling Rectangle could have an effect.

marcsumus:
I'm having the same issue. 

Version 0.50 (55)
MacOS 11.6.1

Nothing in the logs.

rxhanson:
I was able to reproduce this once, but I was unable to determine what caused it, and then it worked after restarting the app :(

If anyone figures out a way to consistently reproduce it, let me know.

marcsumus:
I tried a restart of both the application and the machine and am still not seeing the export file create. Is there anything else you would like me to try? Just looking to help you troubleshoot if I can.

marcsumus:
I'm a java dev not a swift dev, but it feels like this if statement is not resolving to true for some reason perhaps?

https://github.com/rxhanson/Rectangle/blob/9c9c2fc0fe376b95efb8c251825f55dd81f329ff/Rectangle/PrefsWindow/SettingsViewController.swift#L135

rxhanson:
@marcsumus thanks for taking a look. The issue actually appears to be due to macOS not "giving permission to Rectangle to open the save/open panel"; it is unrelated to the actual data being encoded.

othyn:
Appears this issue has crept back up, as its also not working for me. Clicking export and then save in any directory does nothing.

See more: https://github.com/rxhanson/Rectangle/issues/664

azaslavsky:
This is also not working for me as well.  I've tried restarting both the machine and the app as suggested above, to no avail.  I am on Rectangle 0.66 and OSX 12.1 Monterey.

caobo:
I am also having this issue on Rectangle 0.66(72) and macOS 13.2 Ventura.

rxhanson:
I should have a fix out for this in an upcoming release.

[...]

PR #1126, opened 2023-04-02
Fix export with disabled Todo shortcuts

decodism:
Fixes [this](https://github.com/rxhanson/Rectangle/issues/566#issuecomment-1492919023).


rxhanson:
Thanks!

PR #1139, opened 2023-04-24
Added Lithuanian language translation

firebotQL:
Added Lithuanian localisation.

### DIFF
--- a/Rectangle/lt.lproj/Main.strings
+++ b/Rectangle/lt.lproj/Main.strings
@@ -312,7 +312,7 @@
 "YEy-JH-Tfz.title" = "Rasti ir pakeisti…";
 
 /* Class = "NSMenuItem"; title = "\tDefault"; ObjectID = "YGs-j5-SAR"; */
-"YGs-j5-SAR.title" = "\tПо умолчанию";
+"YGs-j5-SAR.title" = "\tPagal nutylėjimą";
 
 /* Class = "NSMenuItem"; title = "Preferences…"; ObjectID = "YRC-4a-xGg"; */
 "YRC-4a-xGg.title" = "Nustatymai…";
@@ -688,7 +688,7 @@
 "N2U-pY-CLq.title" = "Laikyti pasirinktą programą nuolat matomą dešinėje pagrindinio ekrano pusėje";
 
 /* Class = "NSTextFieldCell"; title = "1. Bring your chosen todo application frontmost"; ObjectID = "qze-7p-m1X"; */
-"qze-7p-m1X.title" = "1. Разместите выбранное в качестве Todo приложение на переднем плане";
+"qze-7p-m1X.title" = "1. Pirmame plane padėkite pasirinktą Todo programą";
 
 /* Class = "NSTextFieldCell"; title = "2. In the Rectangle menu, select\n\"Use [Application] as Todo App\""; ObjectID = "xNi-9K-fnJ"; */
 "xNi-9K-fnJ.title" = "„Rectangle“ meniu pasirinkite „Naudoti [Programą] kaip Todo programą“";
--- a/Rectangle/lt.lproj/Main.strings
+++ b/Rectangle/lt.lproj/Main.strings
@@ -673,7 +673,7 @@
 
 "Use frontmost.app as Todo App" = "Naudoti pirmame plane esantį langą kaip Todo programą";
 "Enable Todo Mode" = "Įjungti Todo programos režimą";
-"Perdėlioti Todo" = "Atnaujinti Todo lango padėtį";
+"Reflow Todo" = "Atnaujinti Todo lango padėtį";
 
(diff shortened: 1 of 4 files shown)


firebotQL:
If I've missed something, did incorrectly or there is any other feedback please let me know.

Rectangle/lt.lproj/Main.strings
```diff
@@ -0,0 +1,749 @@
...
+"Xc8-Sm-pig.title" = "Kairė pusė";
+
+/* Class = "NSMenuItem"; title = "Find…"; ObjectID = "Xz5-n4-O0W"; */
+"Xz5-n4-O0W.title" = "Rasti…";
+
+/* Class = "NSMenuItem"; title = "Find and Replace…"; ObjectID = "YEy-JH-Tfz"; */
+"YEy-JH-Tfz.title" = "Rasti ir pakeisti…";
+
+/* Class = "NSMenuItem"; title = "\tDefault"; ObjectID = "YGs-j5-SAR"; */
+"YGs-j5-SAR.title" = "\tПо умолчанию";
```
rxhanson, line 315: Just double-checking if this is correct since it's Cyrillic.

rxhanson:
Thanks for contributing! There's just one comment from reading through it. Also, the file has to be registered in Xcode. I can do that if you don't know how.

Rectangle/lt.lproj/Main.strings
firebotQL: Yes, sorry. Just found out I missed translations in 2 places. Here is the regex to use to find if there are any Cyrillic characters left: [\u0400-\u04FF]. Committing an update shortly.

firebotQL:
> Thanks for contributing! There's just one comment from reading through it. Also, the file has to be registered in Xcode. I can do that if you don't know how.

Let me try to do it. Installing xcode atm. Is that how you do it?



> Thanks for contributing! There's just one comment from reading through it. Also, the file has to be registered in Xcode. I can do that if you don't know how.

I will try add the file from Xcode when I return from work. Also, I've noticed that other translation contributions have included translations for RectangleLauncher/[language].lproj/Main.strings. Would you like me to work on that as well? Please let me know if this is something you'd also like me to complete.

firebotQL:
@rxhanson  I've added RectangleLauncher translations and included both localizations through Xcode. However, could you please take a look at the Main.storyboard file? I'm not certain why the heights and widths have changed in some of the elements. If they are incorrect, I'd appreciate it if you could fix them. I don't believe I've made any mistakes with any of the items, and it's possible that the newer Xcode made these changes by default. But not sure...

Rectangle/lt.lproj/Main.strings
```diff
@@ -0,0 +1,749 @@
...
+
+/* Class = "NSTextFieldCell"; title = "Todo application width"; ObjectID = "6e0-ji-qXw"; */
+"6e0-ji-qXw.title" = "Todo programos lango plotis";
+
+/* Class = "NSTextFieldCell"; title = "Reflow Todo shortcut"; ObjectID = "Fx0-sm-DrT"; */
+"Fx0-sm-DrT.title" = "Karštasis klavišas Todo lango padėties atnaujinimui";
+
+"Use frontmost.app as Todo App" = "Naudoti pirmame plane esantį langą kaip Todo programą";
+"Enable Todo Mode" = "Įjungti Todo programos režimą";
+"Perdėlioti Todo" = "Atnaujinti Todo lango padėtį";
```
rxhanson, line 676: Looks like a few of the English keys got translated into Lithuanian, here and at the end. Would you mind making sure that the correct English keys are in? I'd give it a go myself, but sometimes these can move around between the different translation files and it's probably easier for you to check.

rxhanson:
Thanks for the follow up. The launcher translation wasn't necessary (sorry, I need to update documentation around this and/or rip them all out). Xcode tends to make those size adjustments from time to time, and I'll make sure things look fine after a merge.

rxhanson:
@firebotQL Just following up, in case you didn't catch the comment after your latest revision:

> Looks like a few of the English keys got translated into Lithuanian, here and at the end. Would you mind making sure that the correct English keys are in? I'd give it a go myself, but sometimes these can move around between the different translation files and it's probably easier for you to check.

No rush on a reply, I don't have a release planned for a little while.

firebotQL:
> @firebotQL Just following up, in case you didn't catch the comment after your latest revision:
> 
> > Looks like a few of the English keys got translated into Lithuanian, here and at the end. Would you mind making sure that the correct English keys are in? I'd give it a go myself, but sometimes these can move around between the different translation files and it's probably easier for you to check.
> 
> No rush on a reply, I don't have a release planned for a little while.

Yes, sorry. Did mean to look into it yesterday, but didn't have chance. Will try to address your comments today.

firebotQL:
@rxhanson tagging you here as pushed the change yesterday. let me know if anything else is amiss.

rxhanson:
Thanks!

PR #1220, opened 2023-07-30
Stage Manager: fix multi-screen

decodism:
Tries to fix multi-screen issues with Stage Manager.


rxhanson:
Thanks! I was actually walking this same path and was working on an almost identical version of your `isStageStripVisible()`. This pull request looks good to me, and I'll do some testing on my side before merging.

rxhanson:
Looks good! I'll push out a release in the next day or so.

PR #1245, opened 2023-08-19
Remove information about Cocoapods and replace it with SPM

KORraNpl:
Going through the docs, I noticed some outdated information about Cocoapods being used.


rxhanson:
Thanks!

PR #1252, opened 2023-09-04
Localisation to Norwegian (Nynorsk)/NN, added untranslated files for a possible Norwegian (bokmål)/NB translation

snomos:
Didn't manage to build, so haven't been able to test how well the translations fit the available space in the UI. Would it be possible to get a prerelease build that I could test?

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -245,6 +245,10 @@
 		9824704A22B189250037B409 /* BestEffortWindowMover.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = BestEffortWindowMover.swift; sourceTree = "<group>"; };
 		9824705022B28D7A0037B409 /* LeftRightHalfCalculation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LeftRightHalfCalculation.swift; sourceTree = "<group>"; };
 		983BBD6E253B609D000D223E /* FootprintWindow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FootprintWindow.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


rxhanson:
Thanks, I'll update the localizations in Xcode and let you know when there's a build to look at.

rxhanson:
Sorry for taking so long to get back to this. Here's the link to the build that was generated:
https://github.com/rxhanson/Rectangle/actions/runs/6162760673?pr=1252

Note that it's not signed, so you'll have to bypass gatekeeper to install it.

Let me know if you need any assistance here.

snomos:
Thanks for the build, I got pass gatekeeper.

Unfortunately I have the same issue that I got when I tried building it myself: it starts, and then present me with the dialog that says I need to authorise Rectangle in System preferences. I do, and restart Rectangle, but whatever I do, it can't come pass this point - even if authorised, it still tells me to authorise. The good news is that the dialog is in Norwegian, the bad thing is I can't actually run the app beyond this point.

I assume there is a settings file or something that has been corrupted or something, any advice on how to clean up would be greatly appreciated.

rxhanson:
Sure, so what's going on here is that the accessibility permissions are associated with the "fingerprint" of the signed app and not the unsigned app.

In the case of Xcode, if you leave it configured without a developer id then I think it will require you to "reset" the accessibility permissions every time you build and run the app. I need to note this in the Readme.

You can reset the accessibility permissions by the following:
1. Quit Rectangle.
2. Go to System Settings -> Privacy & Security ->  Accessibility
3. Disable Rectangle with the toggle
4. Remove Rectangle with the minus button
5. Relaunch the unsigned version and add the accessibility permissions as prompted.

Thanks again, sorry this is an annoying obstacle.

snomos:
Thanks a lot for the help, removing Rectangle from the preferences with the minus button did the trick - hadn't tried that.

The Norwegian nynorsk localisation is good now, after I added a few small edits.

snomos:
Both NN and NB localisations done now, feel free to include them in the next release or whenever suits you.

Thanks for the help!

rxhanson:
Thanks! This will be in the next release.

PR #1268, opened 2023-09-21
Filter level when getting window by location

decodism:
Tries to fix https://github.com/rxhanson/RectanglePro-Community/issues/353

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -297,7 +297,7 @@ extension AccessibilityElement {
     }
     
     private static func getWindowInfo(_ location: CGPoint) -> WindowInfo? {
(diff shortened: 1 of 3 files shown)


rxhanson:
Thanks! This is a good way to handle this that I didn't really think through initially.

rxhanson:
I'm going to push out a Rectangle Pro beta with your change, since there are more users reporting there, and will follow up with a Rectangle release depending on the response. Even if there are still some apps that cause issues, this is definitely a step in the right direction.

rxhanson:
I've been considering reducing that window level threshold from 1000 to 20. This way it will ignore the Notification Center (level 23). Do you know of any windows above level 20 that would be valid for drag to screen edge snapping?

rxhanson:
@decodism, thanks! I think this might be the best option I have at the moment so I'm going to go with it.

PR #1277, opened 2023-10-13
Localization update

VChristinne:
Updated old localization format to string catalog.


rxhanson:
Thanks! Sorry for the delay - I've been away from my Mac for a bit and will be reviewing this soon.

rxhanson:
Finally got a chance to take a look. I think something might have to be adjusted. The localization for the helper app appears to be properly loaded in, but the localization for the main app doesn't. There's also a "Localizable" file that maybe needs to be checked in? Here's what I see in the sidebar for the main app: 

image

Here's what the launcher app looks like (I'm assuming this is the desired outcome):

image

Thanks again for the contribution!

VChristinne:
I was supposed to delete that red location file, but apparently I forgot. I'll fix it and check out this main app problem.

VChristinne:
Screenshot 2023-10-23 at 12 29 16 PM


The main app is already here, I'll compare the two to find out why the main one isn't 100%

VChristinne:
@rxhanson I've updated the PR, please check it out

rxhanson:
Sorry, one more thing before merging: looks like we can delete all of the .lproj in `RectangleLauncher`, correct? I'm assuming that Xcode took care of the migration for us; it's interesting that the RectangleLauncher ones have green check marks even if they're not actually translated (not a big deal, just curious about what it's doing).

rxhanson:
I went ahead and deleted all of the lproj in Rectangle Launcher (except the mul.lproj). I'll go ahead and merge this, as it looks to be working fine to me.

PR #1299, opened 2023-12-05
Fix typos in TerminalCommands

FireFingers21:


TerminalCommands.md
```diff
@@ -103,7 +103,7 @@ defaults write com.knollsoft.Rectangle todo -int 1
 
 ## Only allow drag-to-snap when modifier keys are pressed
 
-Modifier key values can be ORed together.
```
rxhanson, line 106: "ORed" is the correct term here. We can replace it with something like "Modifier key values can be combined by OR operation"

rxhanson:
Thanks for contributing! One comment inline.

FireFingers21:
Gotcha, I'm not sure if this is what you meant, but modifer keys can be combined if you add their sums together, so I've changed the wording to reflect that

FireFingers21:
I've tested several different combinations of modifier keys and adding the sum to combine them does actually work. The same applies when creating macOS's built-in "Hot Corners" from Terminal, which I have also done and re-verified.

rxhanson:
Ah, you're right! There's no bit overlap, so it's the same as addition in this case.

PR #1300, opened 2023-12-06
Fix title bar element detection

### LINKED ISSUE #1293: Double-click window title bar to maximize/restore issue with Outlook.app search box.
macOS version: 13.6.1 (22G313) Ventura
Rectangle version: v0.74 (80)
Outlook for Mac version: 16.79.1 (23111718)

When you double click on text in the Outlook.app search box, the "Double-click window title bar to maximize/restore" feature is triggered and the Outlook.app window is resized. 

It seems Outlook utilises the window title bar to show the search box, maybe you can exclude these double clicks as this is used to select all the text in this box.

decodism:
Fixes #1293. Search in children rather than just using the API.

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -172,6 +172,22 @@ class AccessibilityElement {
         return elements
     }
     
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks again!

PR #1321, opened 2024-01-14
Added italian translation

lz90:
Hello from Italy,
this is my first pull request ever!
Double checked the translations with the ones on my MacBook


rxhanson:
Hi, thanks for contributing. Sorry about this, but that file actually doesn't need translation and I thought I had deleted it. If you load the Xcode project, there is only one file that keeps track of all of the localizable elements, and it's at `Rectangle/Main/Main (Strings)`. Italian is currently 100% localized there, but I'm always open to merging improvements if they exist.

I'll go ahead and clean up the unnecessary files for the next release.

PR #1370, opened 2024-04-24
Double/Halve the Width/Height of a window

isaacy2012:
Addresses https://github.com/rxhanson/Rectangle/discussions/1369A

In the doubling case, the direction dictates which way the window will grow. In the halving case, the direction dictates which half the window will remain in.

So, doubling the width right, then halving the width left will restore the window back to the original size and location.

I am unable to test whether whether doubling the width/height allows for spanning across multiple screens, as I don't have a multi-monitor set up on my personal machine. To test this, please first enable support for windows spanning multiple monitors by *toggling off*: 

`System Settings -> Desktop & Dock -> Mission Control -> Displays have separate Spaces`:

Screenshot 2024-04-24 at 9 55 13 PM


rxhanson:
Thanks! The code changes look good. I tested it out using my MacBook Pro and an external display, and unfortunately it behaves a little unexpectedly. I tried starting a window on the left display and doubled width to the right. When the window reaches a size that you would expect to continue onto the next display, it keeps the window on the same display and doubles the width, moving the leftmost side of the window offscreen instead of the rightmost side of the window onto the next display.

isaacy2012:
Hi, thanks for checking that. I realized I can connect my iPad via sidecar so I was able to test and fix it (at least for small windows as the iPad is quite small 😊). If you are able to double-check it on an actual multi-monitor setup that would be great!

rxhanson:
Ah, yes, now I remember that the BestEffortWindowMover had that restriction. All set. I'll be pushing out a release this weekend!

PR #1381, opened 2024-05-04
Update Japanese translation

ryonakano:
Edited using XCode 15.3.

### Changes Summary
- Add translations for new or needs-review strings
- Use Hankaku slash instead of Zenaku one because macOS apps like Notes uses Hankaku one
- Use the translation of "Spelling and Grammar" used in macOS


rxhanson:
Thanks!

PR #1387, opened 2024-05-09
Add separate gap for notched display

Tmate6:
The top gap is calculated from the notch. This means windows overlap custom menubars when having external displays, as the gap is either too small on the external or huge on the built in display. Tried to explain this coherently, hope it makes sense

### DIFF
--- a/Rectangle/ScreenDetection.swift
+++ b/Rectangle/ScreenDetection.swift
@@ -156,8 +156,18 @@ extension NSScreen {
         newFrame.origin.x += Defaults.screenEdgeGapLeft.cgFloat
         newFrame.origin.y += Defaults.screenEdgeGapBottom.cgFloat
         newFrame.size.width -= (Defaults.screenEdgeGapLeft.cgFloat + Defaults.screenEdgeGapRight.cgFloat)
-        newFrame.size.height -= (Defaults.screenEdgeGapTop.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
-
+        
+        if #available(macOS 12.0, *) {
+            if self.safeAreaInsets.top > 0, Defaults.screenEdgeGapTopNotch.value != 0 {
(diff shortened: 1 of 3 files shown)


rxhanson:
Thanks for contributing. I'm trying to think through if it makes sense to automatically take the notch into account for screenEdgeGap settings, and I'm guessing that it would be subjective and probably not make sense.

Rectangle/ScreenDetection.swift
```diff
@@ -156,8 +156,18 @@ extension NSScreen {
...
+            if self.safeAreaInsets.top > 0, Defaults.screenEdgeGapTopNotch.value != 0 {
+                newFrame.size.height -= (Defaults.screenEdgeGapTopNotch.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
+            } else {
+                newFrame.size.height -= (Defaults.screenEdgeGapTop.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
+            }
+            
+        } else {
+            newFrame.size.height -= (Defaults.screenEdgeGapTop.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
+        }
+
```
rxhanson, line 170: Can we condense this down to

```
        if #available(macOS 12.0, *), self.safeAreaInsets.top > 0, Defaults.screenEdgeGapTopNotch.value != 0 {
            newFrame.size.height -= (Defaults.screenEdgeGapTopNotch.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
        } else {
            newFrame.size.height -= (Defaults.screenEdgeGapTop.cgFloat + Defaults.screenEdgeGapBottom.cgFloat)
        }
        
```

Tmate6:
Yes, I originally thought the same, but I think it is more flexible this way.

rxhanson:
Thanks!

PR #1391, opened 2024-05-21
docs: update README to mention `--zap` for Homebrew uninstall

### LINKED ISSUE #1246: Config is not synced from ~/Library/Application Support/Rectangle/RectangleConfig.json

macOS version: 14.0
Rectangle version: 0.70 (76)

The documentation states:

> Upon launch, Rectangle will load a config file at `~/Library/Application Support/Rectangle/RectangleConfig.json` if it is present and will rename that file with a time/date stamp so that it isn't read on subsequent launches.

However, I've made some changes to the shortcuts, exported the `.json` config from Rectangle, symlinked it to `~/Library/Application Support/Rectangle/RectangleConfig.json` and then reloaded Rectangle. It appears that everytime I restart Rectangle it loses all of my config/preferences.

jacobbednarz:
Adds some pointers to the documentation to help those with Homebrew uninstallation where `--zap` is required otherwise, they may see issues with configuration persisting when upgrading from older versions.

Closes #1246

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,12 +127,27 @@ That file can be backed up or transferred to other machines.
 
 If you are using Rectangle v0.44+, you can also use the import/export button in the Preferences pane to share to your preferences and keyboard shortcuts across machines using a JSON file.
 
+> [!NOTE]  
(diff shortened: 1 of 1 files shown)


rxhanson:
thanks!

PR #1425, opened 2024-07-18
Update simplified chinese translation

quan-goh:
Added translations for new or needs-review strings.

Needs Review item:
For "Dismiss" there's multiple translation and I'm not entirely certain on the context so I kept it as Needs Review.

Currently I added it as "忽略" which means either disregard or ignore(as in ignoring the warning/alert/notification). 

But if it should mean :
1. cancel - as in cancelling an action, "取消“ should be used;
2. close- as in closing the dialog/notification popup, "关闭" would be more appropriate.

### DIFF
--- a/Rectangle/mul.lproj/Main.xcstrings
+++ b/Rectangle/mul.lproj/Main.xcstrings
@@ -13779,7 +13779,7 @@
       "localizations" : {
         "zh-Hans" : {
           "stringUnit" : {
-            "state" : "needs_review",
+            "state" : "translated",
             "value" : "忽略"
           }
         }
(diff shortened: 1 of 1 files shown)


rxhanson:
Thanks! The scenario where this is being used is in a popup window with three options. In macOS 15, the window snapping built-in to macOS will conflict with Rectangle's, so the options are to turn off the macOS snapping, turn off the Rectangle snapping or dismiss the popup window without disabling either. 

It seems like this might translate best to "disregard/ignore", but I could also see "close" as being acceptable. What do you think?

quan-goh:
I think in this case 忽略 as in "disregard/ignore" would fit better.

PR #1426, opened 2024-07-19
Check if window snapping is enabled prior to prompting user to disable it

strategiconquest:
I forgot to add a check for this in my last pull request.


rxhanson:
Thanks for catching that & submitting a fix!

Issue #1353, opened 2024-03-27
Use Preferences Toolbar Style for the Settings Window


alexkaessner:
The settings window uses a toolbar style that isn’t recommended by [Apple’s HIG](https://developer.apple.com/design/human-interface-guidelines/toolbars#Platform-considerations).

The common style designed for settings is the [`NSWindowToolbarStylePreference`](https://developer.apple.com/documentation/appkit/nswindowtoolbarstyle/nswindowtoolbarstylepreference) (more at [this WWDC video](https://developer.apple.com/wwdc20/10104?time=423)).

Also a bit cumbersome is the fact that the window size changes so much between the different panes. That leads to the issue that you have to recalibrate your cursor every time. Other apps like Safari or Mail use the same width for all panes and center the content inside the window.

https://github.com/rxhanson/Rectangle/assets/6615076/9714b585-064d-4544-b758-3d768c2864e5

macOS version: 14.3.1
Rectangle version: 0.76 (82)

rxhanson:
Thanks for reporting. I agree, I've always been annoyed by the resizing mechanism here. I don't have the bandwidth to update this at the moment, but would welcome a good pull request that fixes this.

PR #1428, opened 2024-07-24
Set settings screen size to same minimum width

edwinish:
Hello,

Submitting this pull request in reference to this bug.

https://github.com/rxhanson/Rectangle/issues/1353#issue-2210442523

It does not update to the new way Apple recommends. It addresses "Also a bit cumbersome is the fact that the window size changes so much between the different panes.".

I just set the minimum size of the "Settings View Controller Scene", "Snap Area View Controller" to 732, which is the size of the biggest "Prefs View Controller".

Screenshot of settings I changed.
settings-screenshot

Screencapture of settings.
https://github.com/user-attachments/assets/da77c40d-aefa-4f84-9f39-ff1f903ee65c

Let me know if you have any questions or concerns.

### DIFF
--- a/Rectangle/Base.lproj/Main.storyboard
+++ b/Rectangle/Base.lproj/Main.storyboard
@@ -2,7 +2,7 @@
 <document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="32700.99.1234" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22690"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22689"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <scenes>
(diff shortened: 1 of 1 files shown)


rxhanson:
Thanks for working on this! In its current state, I'm not quite on board to merge. This is mainly because the pane on the gear tab is not ready for the increased width, but also for some reason I kind of dislike the additional width on the snapping tab, too. It can be nontrivial to get everything looking great with this type of adjustment, and maybe it's finally worth considering putting all of the shortcuts in one column. Honestly any route that feels desirable to me is enough effort that I had considered waiting until a time when I could rewrite it in SwiftUI and make it look prettier in general, with an updated shortcut recorder (which of course snowballs into a whole lot of effort and part of why the UI has stayed the way it is). 

An interim approach that I would be more inclined to merge would be re-doing the toolbar to position the toolbar items next to the red-yellow-green buttons (using a flexible spacer to the right). We wouldn't get the fixed width, but the more glaring issue of the toolbar items moving around would be fixed.

edwinish:
Thank you for the feed back.

I agree I see the concern with the issue on the sizing of the gear tab.

I will attempt the other interim fix and see how that goes.

edwinish:
Apologies, I spent sometime on it but I am not too familiar with with the storyboard. Or not sure if I can align the tab view to the left side. It seems the tab view is limited but I may be wrong.

I will close this and maybe look for other bug I might be able to help solve.

Thank you for your time.

rxhanson:
No worries, it's not a trivial change (which is why I haven't already made it myself :)

I appreciate your time regardless.

PR #1163, opened 2023-06-02
MultiWindowManager.cascadeActiveAppWindowsOnScreen: resize activeApp …

rudifa:
…windows to the size of the first

PR #1146, opened 2023-05-06
add action cascadeActiveApp; add cascadeActiveAppWindowsOnScreen

rudifa:
Changes:
- enum `WindowAction`: add `case cascadeActiveApp` and update functions that handle the cases
- `class MultiWindowManager`: add `cascadeActiveAppWindowsOnScreen()` and a call from `execute()`
- `TerminalCommands.md`: add section `## Add additional tiling and cascading commands`

@rxhanson @dirondin 

Please review the proposed changes and test them.

I tested the debug build on my Intel Mac (2017), and it works to my satisfaction.

Let me know if I missed anything. Kudos to both of you for the clean and understandable code, easy to extend.

@rudifa

PR #1431, opened 2024-07-27
Add CascadeActiveAppWindows refinements

rudifa:
This PR proposes refinements to the command `cascadeActiveApp` (published in `rectangle v0.69` from PR #1146 and #1163)

- before: the cascade goes to the top left quadrant
- now: the cascade goes to the quadrant in which the center of the active window is found, so that the user can have up to 4 different apps cascaded

- before: the size of the active (first) window is enforced onto the other windows in the cascade
- now: same, but the size width and height are constrained so that all cascaded windows remain within the screen

### DIFF
--- a/Rectangle/MultiWindow/MultiWindowManager.swift
+++ b/Rectangle/MultiWindow/MultiWindowManager.swift
@@ -113,18 +113,18 @@ class MultiWindowManager {
         }
     }
 
-    private struct CascadeActiveAppWindows {
-        var right = false
-        var bottom = false
-        var n = 0
-        var size = CGSizeZero
+    private struct CascadeActiveAppParameters {
+        let right: Bool
+        let bottom: Bool
+        let numWindows: Int
+        let size: CGSize
 
-        init(windowFrame: CGRect, screenFrame: CGRect, n: Int, size: CGSize, delta: CGFloat) {
+        init(windowFrame: CGRect, screenFrame: CGRect, numWindows: Int, size: CGSize, delta: CGFloat) {
             right = windowFrame.midX > screenFrame.midX
             bottom = windowFrame.midY > screenFrame.midY
-            self.n = n
-            let m = CGSize(width: screenFrame.width - CGFloat(n - 1) * delta, height: screenFrame.height - CGFloat(n - 1) * delta)
-            self.size = CGSize(width: min(size.width, m.width), height: min(size.height, m.height))
+            self.numWindows = numWindows
+            let maxSize = CGSize(width: screenFrame.width - CGFloat(numWindows - 1) * delta, height: screenFrame.height - CGFloat(numWindows - 1) * delta)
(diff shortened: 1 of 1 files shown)


rudifa:
Example screenshot:

Screenshot 2024-07-27 at 20 58 50

rxhanson:
Thanks for continuing to improve your work here! I understand the changes you're making, and this corner of the code is mainly only touched by you, but if you don't mind let's make the names of everything more readable. I'll comment inline.

Rectangle/MultiWindow/MultiWindowManager.swift
```diff
@@ -113,6 +113,21 @@ class MultiWindowManager {
         }
     }
 
+    private struct CascadeActiveAppWindows {
```
rxhanson, line 116: Maybe change to `CascadeActiveAppParameters`

Rectangle/MultiWindow/MultiWindowManager.swift
```diff
@@ -113,6 +113,21 @@ class MultiWindowManager {
...
+        var bottom = false
+        var n = 0
+        var size = CGSizeZero
+
+        init(windowFrame: CGRect, screenFrame: CGRect, n: Int, size: CGSize, delta: CGFloat) {
+            right = windowFrame.midX > screenFrame.midX
+            bottom = windowFrame.midY > screenFrame.midY
+            self.n = n
+            let m = CGSize(width: screenFrame.width - CGFloat(n - 1) * delta, height: screenFrame.height - CGFloat(n - 1) * delta)
+            self.size = CGSize(width: min(size.width, m.width), height: min(size.height, m.height))
```
rxhanson, line 127: n and m here take some reasoning to figure out what they mean. Can you think of better names for them?

Rectangle/MultiWindow/MultiWindowManager.swift
```diff
@@ -127,18 +142,19 @@ class MultiWindowManager {
...
+        var caaw: CascadeActiveAppWindows?
 
-        // move the first to become the last
-        if let first = filtered.first, hasFrontWindowPid(first) {
+        if let first = filtered.first {
+            // move the first to become the last (top)
             filtered.removeFirst()
             filtered.append(first)
-            firstSize = first.size
+            caaw = CascadeActiveAppWindows(windowFrame: first.frame, screenFrame: screenFrame, n: filtered.count, size: first.size!, delta: delta)
```
rxhanson, line 152: If `CascadeActiveAppWindows` becomes `CascadeActiveAppParameters`, then we could change `caaw` to something like `cascadeParameters`

rudifa:
@rxhanson Thank you for the suggestions to make the code more readable, which I now applied.

rxhanson:
Looks good!

rudifa:
Thank you!

PR #1470, opened 2024-09-18
Update Japanese translation

ryonakano:
Edited using XCode 16.0.

### Changes Summary
- Add translations for new or needs-review strings
- Use Hankaku exclamation mark instead of Zenaku one because macOS apps like Hints uses Hankaku one


rxhanson:
Great, thanks!

PR #1476, opened 2024-09-22
Update Main.xcstrings (DE localization)

codecantata:
Improvements to the German localization. Some missing translations in connection with the todo mode have been translated.


rxhanson:
Thanks!

PR #1481, opened 2024-10-05
docs: Add missing `cascade-active-app` and `center-prominently` window commands

5HT2:
You can find references to the both of these inside of `./Rectangle/WindowAction.swift`, as well as inside of `./TerminalCommands.md`


rxhanson:
Thanks!

PR #1482, opened 2024-10-07
Handle additional tiling setting in macOS 15.1

strategiconquest:
In the macOS 15.1 beta, there's an additional setting for the top snap area that needs to be checked by Rectangle.


rxhanson:
Looks great! Thanks for your continued work on this 👍

PR #1504, opened 2024-10-28
Add commands for constrained resize

nikhilweee:
I love the resize commands in rectangle but often I wish there was a way to only resize in one dimension. This PR is an attempt to add those commands. There is a related discussion in https://github.com/rxhanson/Rectangle/discussions/387 which is the inspiration for this PR.

This PR adds two new window actions, `largerWidth` and `smallerWidth`, which are intended to make the window width larger / smaller without changing its height. (Technically, `largerHeight` can be achieved by `larger` followed by `smallerWidth` and `smallerHeight` can be achieved by `smaller` followed by `largerWidth`, so those are omitted.)

I have defined the two new actions in `WindowAction.swift`. I have also made changes in `ChangeSizeCalculation.swift` for the new actions. The default shortcuts for the two new commands are defined as follows:

```
case .largerWidth: return Shortcut( ctrl|alt|shift, kVK_ANSI_Equal )
case .smallerWidth: return Shortcut( ctrl|alt|shift, kVK_ANSI_Minus )
```

That being said, I do not have prior experience building macOS apps or using XCode, so I have not been able to test the changes. When building the project on XCode, the build succeeds and I get a prompt to "Authorize Rectange" which asks me to open system settings and enable "Rectangle.app", but it's already enabled. Any help would be appreciated.


rxhanson:
Thanks for contributing. I haven't looked at your change yet, but wanted to answer this:

> That being said, I do not have prior experience building macOS apps or using XCode, so I have not been able to test the changes. When building the project on XCode, the build succeeds and I get a prompt to "Authorize Rectange" which asks me to open system settings and enable "Rectangle.app", but it's already enabled. Any help would be appreciated.

There's a bug (at least I think it's a bug?) with accessibility privileges that makes developing without a paid developer ID a pain here. You will need to 
1. Quit Rectangle
2. Remove Rectangle from the Security & Privacy Accessibility System Setting. (You might have to toggle it off first).
3. Launch Rectangle from Xcode and provide privileges as prompted.

You'll have to do this every time you make a change :(

I think a long time ago I had heard perhaps you can give Xcode accessibility privileges to get around this, but that didn't work for me then and I don't think it would make a difference.

nikhilweee:
Hey @rxhanson Thanks so much for the tip. I was able to build and run Rectangle using XCode and I can confirm that the new commands work as expected. Please feel free to take a look.

At the moment I have hardcoded the keyboard shortcuts for the new commands as posted before. I don't know if there is a way for someone (the user) to change the shortcuts later.

rxhanson:
Glad that you've got things working! You shouldn't have to do anything extra to allow these actions to be configured via Terminal, see an example: https://github.com/rxhanson/Rectangle/blob/main/TerminalCommands.md#add-an-extra-centering-with-prominence-command

I'll give things a thorough test soon. In the meantime, things look good; the only things that jumped out at me the code are:
1. Some of the indentation in `ChangeSizeCalculation` needs to be adjusted.
2. We'll want to remove those hardcoded default shortcut values.

nikhilweee:
@rxhanson I fixed indentation, removed default shortcuts and added some documentation.

rxhanson:
Thanks!

PR #1505, opened 2024-10-28
Handle ignore unignore URL actions

bric3:
Adds the support for

* `open -g "rectangle://execute-action?name=ignore-app"`
* `open -g "rectangle://execute-action?name=unignore-app"`

### DIFF
--- a/Rectangle/AppDelegate.swift
+++ b/Rectangle/AppDelegate.swift
@@ -552,18 +552,23 @@ extension AppDelegate {
             prevActiveApp?.activate()
         }
         DispatchQueue.main.async {
+            
             func getUrlName(_ name: String) -> String {
                 return name.map { $0.isUppercase ? "-" + $0.lowercased() : String($0) }.joined()
             }
-            func getAppBundleId(_ components: URLComponents) -> String? {
-                let appBundleId = (components.queryItems?.first { $0.name == "app-bundle-id" })?.value
-                Logger.log("The app Bundle id : \(appBundleId ?? "")")
-                return if appBundleId != nil && appBundleId!.isEmpty  {
-                    appBundleId
-                } else {
-                    ApplicationToggle.frontAppId
-                }
+            
+            func extractBundleIdParameter(fromComponents components: URLComponents) -> String? {
+                (components.queryItems?.first { $0.name == "app-bundle-id" })?.value
             }
+            
+            func isValidParameter(bundleId: String?) -> Bool {
+                let isValid = bundleId?.isEmpty != true
+                if !isValid {
+                    Logger.log("Received an empty app-bundle-id parameter. Either pass a valid app bundle id or remove the parameter.")
+                }
+                return isValid
+            }
+            
             for url in urls {
                 guard
                     let components = URLComponents(url: url, resolvingAgainstBaseURL: true),
@@ -578,9 +583,13 @@ extension AppDelegate {
                     let action = (WindowAction.active.first { getUrlName($0.name) == name })
                     action?.postUrl()
                 case ("execute-task", "ignore-app"):
-                    self.applicationToggle.disableApp(appBundleId: getAppBundleId(components))
+                    let bundleId = extractBundleIdParameter(fromComponents: components)
+                    guard isValidParameter(bundleId: bundleId) else { continue }
+                    self.applicationToggle.disableApp(appBundleId: bundleId)
                 case ("execute-task", "unignore-app"):
-                    self.applicationToggle.enableApp(appBundleId: getAppBundleId(components))
+                    let bundleId = extractBundleIdParameter(fromComponents: components)
+                    guard isValidParameter(bundleId: bundleId) else { continue }
+                    self.applicationToggle.enableApp(appBundleId: bundleId)
                 default:
                     continue
                 }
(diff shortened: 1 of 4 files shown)


rxhanson:
Thanks for looking to contribute.

Rather than adding these as window actions, I recommend adding a new url query parameter that will call             `applicationToggle.enableFrontApp()` or `applicationToggle.disableFrontApp()`. That way your changes are only in AppDelegate. The url stuff is at the bottom of AppDelegate.

Maybe something like `rectangle://execute-task?name=ignore-app`?

bric3:
What about adding a query pram for the app bundle id, so that even if IntelliJ is in background for some reason it can execute this action without affecting another app.

rxhanson:
Sure! This looks good to me, but I can hold off on merging if you wanted to add in the bundleId query param.

bric3:
OK I'll do it shortly.

bric3:
Added `app-bundle-id` param for both ignore/unignore.

Rectangle/AppDelegate.swift
```diff
@@ -555,17 +555,36 @@ extension AppDelegate {
...
-                    let name = (components.queryItems?.first { $0.name == "name" })?.value,
-                    let action = (WindowAction.active.first { getUrlName($0.name) == name })
+                    components.path.isEmpty
                 else {
                     continue
                 }
-                action.postUrl()
+                    
+                let name = (components.queryItems?.first { $0.name == "name" })?.value
+                let appBundleId = getAppBundleId(components)
```
rxhanson, line 576: Here's a totally subjective comment, and I'm good with this as-is, but I actually like calling this inline in the cases below like this:
```
self.applicationToggle.disableApp(appBundleId: getAppBundleId(components))
```
No need to change it if you don't want to.

Rectangle/AppDelegate.swift
```diff
@@ -555,17 +555,36 @@ extension AppDelegate {
             func getUrlName(_ name: String) -> String {
                 return name.map { $0.isUppercase ? "-" + $0.lowercased() : String($0) }.joined()
             }
+            func getAppBundleId(_ components: URLComponents) -> String? {
+                let appBundleIdParam = (components.queryItems?.first { $0.name == "app-bundle-id" })
+                // if param not present the behavior is to use front app
+                return if let appBundleIdParam {
```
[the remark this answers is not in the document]
bric3: I pondered on that, but then how to distinguish 

* no `app-bundle-id` query parameter (i.e. the url is `rectangle://execute-task?name=ignore-app`, which would mean the behavior is to use the front app), 
* from an empty value `rectangle://execute-task?name=ignore-app&app-bundle-id=`, in this case the good behavior is to do nothing I think (the `disableApp` only do something if the value is not `null`)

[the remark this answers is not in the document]
rxhanson: In this case, `return appBundleIdParam?.value` would have the identical outcome to the `if let` logic. It's also unusual to return an `if let` and it is not consistent with the style in Rectangle.

If you want to account for an empty string, I don't see any issues with that.

Rectangle/AppDelegate.swift
```diff
@@ -555,17 +555,34 @@ extension AppDelegate {
             func getUrlName(_ name: String) -> String {
                 return name.map { $0.isUppercase ? "-" + $0.lowercased() : String($0) }.joined()
             }
+            func getAppBundleId(_ components: URLComponents) -> String? {
+                let appBundleId = (components.queryItems?.first { $0.name == "app-bundle-id" })?.value
+                return if appBundleId == nil {
```
rxhanson, line 560: Thanks for making this change here. Sorry to beleaguer this, but I figure we might as well get this tidied up. 

* What we have now can also be rewritten as `return appBundleId ?? ApplicationToggle.frontAppId` (hence one of the reasons why you don't typically see `if` come after `return`)
* However, this would still be the same as `return appBundleId` since once we get to `ApplicationToggle`, we would be providing the `ApplicationToggle.frontAppId` at that time in the case of `appBundleId` being nil.
* If we want to account for `app-bundle-id=` in the URL, then we would have to check that `appBundleId != nil && appBundleId.isEmpty` (depending on if we do indeed get back an empty string in that case - I didn't test it).

I'm still ok either way on doing the empty string check, but it still makes the most sense to me to at the very least change this to `return appBundleId` here.

bric3: Sorry, I pushed a change in the train yesterday as a reflex, (if anything happen to the laptop).

I agree with your points; my original goal is indeed the third point.

rxhanson:
I wanted to make sure that I'm not holding things up here - let me know when you're ready for me to take another look (no pressure or anything).

bric3:
@rxhanson Hi, I think I fixed the last open point. And yes I'd like another pass.

Thank you very much!

rxhanson:
Apologies for the delay. I hope you don't mind, but I proposed my expected change directly in your branch in the interest of moving this along. Your last change still allowed for an empty app-bundle-id to result in the frontmost app being ignore/unignored. Feel free to roll back my proposal if this isn't what you were intending.

bric3:
Looks good. Thanks for looking into it.

I didn't write the validation because this empty value has no side effect as far as I know. It's just stored into the "hash set", and since no app can have an empty bundle id (I may be wrong here), this will never be looked up in this "hash set".
That said having correct data in this collection is better.

Issue #1540, opened 2025-01-10
Switch the 4x2 grid with a 2x4 grid for the Eights windows calculation in portrait mode


joaquin-splunk:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 15.1.1 (24B91)
Rectangle version: 3.0.37 (178) (Pro version)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): N/A

When doing the window calculations for most of the "eights" (BottomLeftEighthCalculation being the exception) in portrait mode, it divides the screen by a 4x2 grid. I expected that on portrait mode the grid would switch to 2x4 instead.

Thank you for your time 🙇🏼

PR #1541, opened 2025-01-10
Invert calculation for the "Eights" in portrait mode

joaquin-splunk:
When I was configuring my screen that's on portrait mode with the Eights, the only one that fit my expectation was when it was on the **Bottom Left** corner of my screen. That's when I noticed on the code for [Rectangle/WindowCalculation/BottomLeftEighthCalculation.swift](https://github.com/rxhanson/Rectangle/blob/e481b024e450cef30fb9866b7c905b2f88f4a225/Rectangle/WindowCalculation/BottomLeftEighthCalculation.swift#L45) the divisors are switched for the height and the width, while on the other calculations for the Eights are equal for both landscape and portrait mode.

This PR just switches them for the portrait mode for the other calculations.

### DIFF
--- a/Rectangle/WindowCalculation/BottomCenterLeftEighthCalculation.swift
+++ b/Rectangle/WindowCalculation/BottomCenterLeftEighthCalculation.swift
@@ -42,8 +42,8 @@ class BottomCenterLeftEighthCalculation: WindowCalculation, OrientationAware, Ei
     
     func portraitRect(_ visibleFrameOfScreen: CGRect) -> RectResult {
         var rect = visibleFrameOfScreen
-        rect.size.width = floor(visibleFrameOfScreen.width / 4.0)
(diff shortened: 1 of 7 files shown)


joaquin-splunk:
This is in relation to https://github.com/rxhanson/Rectangle/issues/1540

rxhanson:
Thanks for contributing. I tested this out real quick and it looked like the eighths that go in the center spaces will have to get their origin coordinates adjusted as well?

rxhanson (changes requested):
Figured I'd check back in on this. As noted, it looked to me that we would also need the origins of each rectangle adjusted here as well. I don't mind fixing this, but didn't want to step on any toes if you were still looking to fix it.

joaquin-splunk:
Thanks for checking in. I haven't had the time yet to look into the issue. I might take a look during this weekend or next weekend, unless you get it before then.

Issue #47, opened 2019-12-06
How do I ignore an app?


rxhanson:
To ignore an app:

1. Focus the app that you want to ignore (make a window from that app frontmost). 
1. Open the Rectangle menu and select "Ignore <frontmost app>"
![image](https://user-images.githubusercontent.com/13651296/70309210-420cd480-180d-11ea-8e3c-65b10e95e8f7.png)

To un-ignore an app that you have selected to ignore, simply bring that app frontmost again, open the Rectangle menu, and deselect "Ignore <frontmost app>"

Ignoring an app means that when the app is frontmost, keyboard shortcuts are un-registered from macOS. When the app is no longer frontmost, keyboard shortcuts are re-registered with macOS.

michaeljohnbennett:
the ignore options don't appear for me:

mojave 10.14.6

I'm on Vivaldi (chrome app) and nothing appears:

![image](https://user-images.githubusercontent.com/951378/70314620-f2112a80-186b-11ea-94fc-25b9031d9ba7.png)

Same for Intellij which has a bunch of nav shortcuts that rectangle coincide with.

![Uploading image.png…]()

so something is not showing correctly on the latest version.

rxhanson:
If the Ignore menu item is not shown, it means that Rectangle can't obtain the frontmost app (or the name of the frontmost app). Rectangle relies on macOS to notify it of the frontmost app, so chances are that there's a hiccup macOS side either on registering for the notifications or the notifications themselves.

Try restarting Rectangle first. If you still have the problem then maybe try quitting Rectangle and removing it from System Prefs: Security & Privacy: Privacy: Accessibility. Then run Rectangle and re-add it as prompted. If none of that works, I can add in some logging and we can pinpoint why macOS isn't giving us what we need.

michaeljohnbennett:
ok after a bit of off and on and off again and few restarts its working... 

very strange. the ignore stuff would be good if it was documented somewhere, I had to search through closed issues/tickets to find out if it was there or not.  I went back to spectacle for a day until I knew it was working.

Thanks

:-)

rxhanson:
[...]

PR #1584, opened 2025-04-03
Update README.md - Clarified Ignoring an App

BernhardBln:
Shamelessly copied your explanation from https://github.com/rxhanson/Rectangle/issues/47 and added an additional clarification, after I was wondering what "ignore app" means and how I would un-ignore it.


rxhanson:
Thanks!

PR #1617, opened 2025-07-14
Add center two thirds action

savin000:
I use a 34-inch ultrawide monitor (HUAWEI MateView GT ZQE-CBA), and having a shortcut to limit the screen space like working on a 27-inch display helps me stay focused. In the expanding world of ultrawide monitors, this kind of mode is quickly becoming a must-have.


savin000:
Edited UI Layout:

image

image

rxhanson:
Thanks for contributing! This looks great; the only thing that I have hesitation on is automatically adding a default shortcut, despite it being an obvious choice. There's potential for frustration for people that have configured that shortcut to something else prior to updating Rectangle. In scenarios like this in my other apps, I would typically make it a default for new installs of the app, but not updates. Here in Rectangle, there's not already a great mechanism in place for this scenario. I'm fine going ahead and merging this in and making the change myself to do this, or you can add it if desired and I'll hold off on merging if you want to make it part of this PR.

savin000:
I will try to implement this

savin000:
Since you already have a similar mechanism in other apps, it might be easier for you to implement this on your side. The only approach I’ve come up with is checking the version using something like this:

```
private func isNewInstall() -> Bool {
        let defaults = UserDefaults.standard
        let versionKey = "app_version"
        let currentVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String

        if defaults.string(forKey: versionKey) == nil || defaults.string(forKey: versionKey) != currentVersion {
            defaults.set(currentVersion, forKey: versionKey)
            return true
        }

        return false
    }
```

But to me, it feels a bit clunky. I’m not quite sure how to integrate this with the shortcut logic, as I don’t have much experience with macOS development.

rxhanson:
Thanks for taking a look at implementing it. I'll plan on adding the new install check :)

PR #1621, opened 2025-07-26
Add “Center Three Fourths” action

tgt:
I noticed “Center Two Thirds” in the latest update and thought “Center Three Fourths” would be nice so I could get rid of my approximation using `specified` and I thought it rounded out the default sizes.

I tweaked the icon for “Center Two Thirds” as it was too similar to the “Center Three Fourths” image I've added.

It's localised to en-GB.

Changes based on https://github.com/rxhanson/Rectangle/pull/1617


rxhanson:
Thanks for contributing! At first I was going to force this to be left out of the UI as a Terminal command config only, but I agree, it rounds out the default sizes now.

PR #1628, opened 2025-08-22
Update Korean Translation

zooxop:
This change makes Korean option labels consistent, as all others use the "/" format while this one used "3분의 2".

Rectangle korean translation issue


rxhanson:
Thanks!

PR #1644, opened 2025-09-17
Add option to move resized window to the display under cursor.

fnando:
Divvy has a nice feature where it moves the resized window to the display under cursor. I added a similar feature that's opt-in.

CleanShot 2025-09-17 at 11 40 36

PS: Xcode 26 changed some positioning and is giving a few warnings. I didn't know how you'd like to handle it, so I kept it as it is (you can see the buttons are broken on the image above; same on the initial screen asking for the shortcut profile).

### DIFF
--- a/Rectangle/ScreenDetection.swift
+++ b/Rectangle/ScreenDetection.swift
@@ -33,6 +33,33 @@ class ScreenDetection {
         return UsableScreens(currentScreen: sourceScreen, adjacentScreens: adjacentScreens, numScreens: screens.count, screensOrdered: screensOrdered)
     }
 
+    func detectScreensAtCursor() -> UsableScreens? {
+        let screens = NSScreen.screens
+        guard let firstScreen = screens.first else { return nil }
+
+        if screens.count == 1 {
+            // Delegate to window-based detection for single screen
+            return detectScreens(using: nil)
+        }
+
+        let screensOrdered = order(screens: screens)
+
+        // Get cursor position
+        let mouseLocation = NSEvent.mouseLocation
+
+        // Find screen containing cursor
+        guard let cursorScreen = screens.first(where: { screen in
+            screen.frame.contains(mouseLocation)
+        }) else {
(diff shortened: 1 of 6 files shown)


Rectangle/ScreenDetection.swift
```diff
@@ -33,6 +33,33 @@ class ScreenDetection {
         return UsableScreens(currentScreen: sourceScreen, adjacentScreens: adjacentScreens, numScreens: screens.count, screensOrdered: screensOrdered)
     }
 
+    func detectScreensAtCursor() -> UsableScreens? {
```
rxhanson, line 36: Let's remove all of the comments in this function.

Rectangle/WindowManager.swift
```diff
@@ -64,7 +64,12 @@ class WindowManager {
         if let screen = parameters.screen {
             screens = UsableScreens(currentScreen: screen, numScreens: 1)
         } else {
-            screens = screenDetection.detectScreens(using: frontmostWindowElement)
+            // Use cursor position or window position based on user preference
```
rxhanson, line 67: Let's remove this comment as well, and change this if else to a ternary:

```
screens = Defaults.useCursorScreenDetection.enabled
? screenDetection.detectScreensAtCursor
: screenDetection.detectScreens(using: frontmostWindowElement)
```

rxhanson (commented):
Thanks for contributing!

I'll take a look at those layout issues - thanks for pointing that out.

rxhanson:
I finally had a chance to test this one out, and we'll also want to properly handle next/prev display commands. With your changes, executing those repeatedly will land the window only on the other display and it won't go back. I think my expected behavior would be for those to not use the cursor display as the main display.

rxhanson:
I went ahead and made the changes I had recommended. Because the actions that move to other displays weren't ironed out, I left the checkbox hidden by default (only enabled via Terminal command). If you get a chance to fix up the other display edge cases, make another PR and we can add the checkbox back in.

PR #1649, opened 2025-09-28
Set all settings views to the same width (in a nice looking way)

strategiconquest:
Fixes https://github.com/rxhanson/Rectangle/issues/1353

Different from https://github.com/rxhanson/Rectangle/pull/1428, this change will add whitespace around the existing stack views instead of stretching them the size of the window.


rxhanson:
Perfect! Thanks!

PR #1652, opened 2025-10-06
Discussion #1599, increase/decrease width/height shortcut

thenickygee:
### PR Summary
- Added new button for 'extra settings' in **General** tab
- Added largerWidth and smallerWidth shortcuts (no default values)
- Added 'Width Step' custom integer value input with default of 30px
- Added shortcut icons based on current project asset design styles
- Inspiration from discussion: #1599 

### Requirements from @rxhanson 
- "I'm open to having something like an "even more settings" button" 
- "opens a popover with more settings where this can go"
- "prefer that the button lives in the General tab at first"

### Tests Passing
- Test Suite 'All tests' passed at 2025-10-05
- Executed 2 tests, with 0 failures (0 unexpected) in 0.350 seconds

image


rxhanson:
Looks good! I'll test things out soon, but I don't see any issues with the changes so far.

rxhanson (commented):
Tested it out, and everything works great. A few points worth mentioning:

1. We can change the `widthStepSize` default to have a default value of 30 in Defaults.swift, and then we don't have to repeat the default value in two places. (I know that this wasn't the way it was previously, but it makes sense to have it that way now).
2. It would fit with the style of the shortcuts tab to have the text in the popover right aligned, for both the additional shortcuts and the step size components.
3. We'll need the text that's in the popover to be added to the Main.xcstrings, otherwise it won't get translated when the community looks to translate it.

I don't mind merging as-is and making these changes, but would definitely appreciate it if you went ahead and made those changes. I'll hold off on merging immediately if you would like to make those changes.

Thanks again for contributing!

thenickygee:
I can totally make these needed changes. I'm tracking everything you're requesting and will get my PR updated. 

And of course- happy to be helping out 🤝

thenickygee:
@rxhanson got changes in to address these requests: 

### Changes Summary

- _We can change the widthStepSize default to have a default value of 30 in Defaults.swift..._ 
  - ✅ **Added widthStepSize defaultValue and replaced existing hard coded definitions**

- _It would fit with the style of the shortcuts tab to have the text in the popover right aligned..._ 
  - ✅ **Right aligned text and input content. Center aligned popover title (I felt like the left aligned title looked a bit weird with the new right alignment. Was also thinking about removing the title to the popover altogether but will leave that up to you)**

- _We'll need the text that's in the popover to be added to the Main.xcstrings..._ 
  - ✅ **Added 'extra shortcuts' strings to Main.xcstrings (followed existing patterns) and wrapped the labels with NSLocalizedString.**

### Tests Passing

- Test Suite 'All tests' passed at 2025-10-12
- Executed 2 tests, with 0 failures in 0.338 seconds

1599-new-screenshot

rxhanson:
Excellent! Thanks!

PR #1659, opened 2025-10-28
Liquid Glass App Icon for macOS 26 Tahoe

alexkaessner:
I’ve added a Liquid Glass variant of @gmcusaro’s app icon. ✌️

To keep the legacy icon variant for older systems, I’ve used this undocumented build setting `--enable-icon-stack-fallback-generation=disabled` ([more in the Apple Developer Forums](https://developer.apple.com/forums/thread/787576?answerId=858420022#858420022)). I don’t have an old macOS system at hand right now, but from tests with my own apps, it should work.


rxhanson:
Awesome, thank you!! I'll test things out on an old macOS install and see how that setting works.

PR #1663, opened 2025-11-05
Add toggle between leaving stage manager space and full maximize

benedictschmieder:
## Description

Adds a double-maximize feature that toggles between respecting Stage Manager space and using the full screen width when maximizing windows.

## Behavior

- **First maximize**: Always respects Stage Manager space (leaves room for the Stage Manager strip on the left)
- **Double-maximize**: Pressing the maximize shortcut twice in a row toggles between:
  - Respecting Stage Manager space (default)
  - Ignoring Stage Manager space (full screen width)

## Usage

1. Press `Ctrl + Opt + Enter` once: Window maximizes with Stage Manager space respected
2. Press `Ctrl + Opt + Enter` again immediately: Window expands to full screen width
3. Press `Ctrl + Opt + Enter` again: Returns to respecting Stage Manager space

## Implementation Details

- Adds variable `ignoreStageOnDoubleMaximize` in `Defaults.swift`
- Detects double-maximize in `ShortcutManager.swift` by checking if the last action was also maximize on the same window
- Modifies Stage Manager space calculation in `WindowManager.swift` and `WindowCalculation.swift` to respect the toggle state

### DIFF
--- a/Rectangle/ShortcutManager.swift
+++ b/Rectangle/ShortcutManager.swift
@@ -81,21 +81,6 @@ class ShortcutManager {
             return
         }
         
-        // Handle double-maximize toggle for Stage Manager space
-        if parameters.action == .maximize {
-            guard let windowElement = parameters.windowElement ?? AccessibilityElement.getFrontWindowElement(),
-                  let windowId = parameters.windowId ?? windowElement.getWindowId()
-            else {
-                NSSound.beep()
-                return
-            }
-            
-            if let lastAction = AppDelegate.windowHistory.lastRectangleActions[windowId],
-               lastAction.action == .maximize && lastAction.count >= 1 {
-                Defaults.ignoreStageOnDoubleMaximize.toggle()
-            }
-        }
-        
         // Check if repeat cycles displays
         if Defaults.subsequentExecutionMode.value == .cycleMonitor,
            parameters.action.classification != .size,
--- a/Rectangle/WindowCalculation/WindowCalculation.swift
+++ b/Rectangle/WindowCalculation/WindowCalculation.swift
@@ -65,12 +65,10 @@ struct WindowCalculationParameters {
     let ignoreTodo: Bool
     
     func asRectParams(visibleFrame: CGRect? = nil, differentAction: WindowAction? = nil) -> RectCalculationParameters {
-        let actionToUse = differentAction ?? action
-        let ignoreStage = actionToUse == .maximize && Defaults.ignoreStageOnDoubleMaximize
-        return RectCalculationParameters(window: window,
-                                         visibleFrameOfScreen: visibleFrame ?? usableScreens.currentScreen.adjustedVisibleFrame(ignoreTodo, ignoreStage),
-                                         action: actionToUse,
-                                         lastAction: lastAction)
+        RectCalculationParameters(window: window,
+                                  visibleFrameOfScreen: visibleFrame ?? usableScreens.currentScreen.adjustedVisibleFrame(ignoreTodo),
+                                  action: differentAction ?? action,
+                                  lastAction: lastAction)
     }
     
     func withDifferentAction(_ differentAction: WindowAction) -> WindowCalculationParameters {
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -79,7 +79,6 @@ class Defaults {
     static let stageSize = FloatDefault(key: "stageSize", defaultValue: 190)
     static let dragFromStage = OptionalBoolDefault(key: "dragFromStage")
     static let alwaysAccountForStage = OptionalBoolDefault(key: "alwaysAccountForStage")
-    static var ignoreStageOnDoubleMaximize = false
     static let landscapeSnapAreas = JSONDefault<[Directional:SnapAreaConfig]>(key: "landscapeSnapAreas")
     static let portraitSnapAreas = JSONDefault<[Directional:SnapAreaConfig]>(key: "portraitSnapAreas")
     static let missionControlDragging = OptionalBoolDefault(key: "missionControlDragging")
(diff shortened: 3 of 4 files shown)


benedictschmieder:
**Disclaimer:** The translations were generated using AI and may require review by native speakers to ensure accuracy.

Rectangle/ShortcutManager.swift
```diff
@@ -81,6 +81,21 @@ class ShortcutManager {
             return
         }
         
+        // Handle double-maximize toggle for Stage Manager space
```
rxhanson: Logic specific to a single window action should go into the calculation implementation for the action. In this case, maximize logic should be in the `MaximizeCalculation`.

Rectangle/WindowCalculation/WindowCalculation.swift
```diff
@@ -65,10 +65,12 @@ struct WindowCalculationParameters {
     let ignoreTodo: Bool
     
     func asRectParams(visibleFrame: CGRect? = nil, differentAction: WindowAction? = nil) -> RectCalculationParameters {
-        RectCalculationParameters(window: window,
-                                  visibleFrameOfScreen: visibleFrame ?? usableScreens.currentScreen.adjustedVisibleFrame(ignoreTodo),
-                                  action: differentAction ?? action,
-                                  lastAction: lastAction)
+        let actionToUse = differentAction ?? action
```
rxhanson: Again, this logic should be in the `MaximizeCalculation`.

rxhanson (changes requested):
Thanks for looking to contribute! For this change, the path you'll want to take is to revert these changes and only modify the `MaximizeCalculation` class. From there, you can access the last action to determine if it's a repetition that should get the full maximize vs the maximize leaving space for stage manager. Perhaps it would make sense to alternate between the two (comparing even/odd repetitions?). You can also use some of the StageUtil static variables to determine if stage manager is being used.

Rectangle/mul.lproj/Main.xcstrings
```diff
@@ -47149,181 +47149,181 @@
...
             "state" : "translated",
-            "value" : "Wenn der Bereich zu klein ist, werden die zuletzt verwendeten Programme ausgeblendet."
+            "value" : "Wenn der Bereich zu klein ist, werden die zuletzt verwendeten Programme ausgeblendet. Zweimaliges Drücken der Maximieren-Tastenkombination überschreibt diese Einstellung und nutzt den gesamten verfügbaren Bildschirm."
           }
         },
         "en" : {
           "stringUnit" : {
             "state" : "new",
-            "value" : "If the area is too small, recent apps will be hidden"
+            "value" : "If the area is too small, recent apps will be hidden. Double-pressing the maximize shortcut will override this setting and use all of the available screen space."
```
rxhanson: Recommendation:
Repeat maximize to override and expand fully.

martin-braun:
Would this allow me to force respecting the stage manager space, even if I have only one window?

Right now I boot with only the terminal showing up, I resize it to Rectangle's 'Almost Maximize'. As soon as I open another window the stage manager space appears, but my terminal is covering it, because it was sized without respecting the stage manager's size.

I'm not sure if this PR is the correct scope to ask for the ability for a 3 way toggle (`respect`, `ignore` and `auto`, where `auto` is the current behavior).

I also think that your last selection should be preserved between app launches so that this is not a temporary setting anymore. Reason in my case is that I can simply tell Rectangle to always respect the space of the stage manager, effectively solving my culprit I'm facing.

rxhanson:
@martin-braun The initial change in this pull request was for adjusting the maximize command to ignore the stage manager in subsequent executions. The preferred place to request a change is in [Discussions](https://github.com/rxhanson/Rectangle/discussions). So that I understand better, are you looking for a toggle so that you can disable/enable leaving space for the stage manager without adjusting the slider?

benedictschmieder:
> Thanks for looking to contribute! For this change, the path you'll want to take is to revert these changes and only modify the `MaximizeCalculation` class. From there, you can access the last action to determine if it's a repetition that should get the full maximize vs the maximize leaving space for stage manager. Perhaps it would make sense to alternate between the two (comparing even/odd repetitions?). You can also use some of the StageUtil static variables to determine if stage manager is being used.

Hey,
thanks for your quick review and reply, highly appreciated.
As you can see, I already reverted my changes and am currently working on a better solution which aligns with your recommendations. I am currently facing an issue with multiple screens which I have yet to resolve between my other day to day business. I'll get back to you as soon as I have a working solution.

Kind regards.
Benedict

martin-braun:
> So that I understand better, are you looking for a toggle so that you can disable/enable leaving space for the stage manager without adjusting the slider?

I'm looking for a 3-way toggle without adjusting the slider:

```
[ Yes | Auto | No ] Respect Stage Manager recent apps area
```

- Yes: Never maximize windows fully, even if only one recent app is available
- Auto: Maximize windows full screen or with less width (Width = Maximum width - "Stage Manager recent apps area") <- Current behavior
- No: Always maximize windows fully, effectively ignoring the "Stage Manager recent apps area"

My problem is that my previously maximized window (only one) doesn't shrink when a 2nd recent window appears, so I have to fix the first window again. I have this problem on every boot, because my terminal automatically opens and nothing else:



I end up having to resize my Terminal twice after opening another app, because I don't want my terminal to cover the recent apps area of the Stage Manager. I just wish I could force Rectangle to always respect the "Stage Manager recent apps area", so in regards of this PR, I think a 3 toggle switch that can also be programmed via keymap would solve the limitation for everyone nicely. I hope @benedictschmieder can come up with something, that would be wonderful. :)

rxhanson:
@benedictschmieder No pressure or anything, but I'm going to close this out for now. Feel free to reopen whenever you're ready.

benedictschmieder:
Yeah, absolutely! I forgot that this is still open, sorry.
I've gotten it working but not in a way that I'm satisified with. I'll keep working on it and get back to you when I'm finished.
Thanks for your understanding.

PR #1675, opened 2025-12-14
Added new snap-area for dedicated portrait top/bottom applied to left and right sides.

Da2MinuteNoodle:
# Added new snap-area for dedicated portrait top/bottom applied to left and right sides.

I added this functionality as I find using the "Thirds" option with Top/Bottom in the corners a bit cumbersome...

In my workflow I always prefer having two windows top/bottom on my portrait monitor and felt that a dedicated option would be useful for some.


rxhanson:
Thanks! I'll test this out soon.

rxhanson:
Excellent work, thanks again!

PR #1678, opened 2025-12-22
feat add thirds and two thirds shortcuts

khaledosama999:
### Discussion
https://github.com/rxhanson/Rectangle/discussions/1574

### DIFF
--- a/Rectangle/WindowCalculation/BottomVerticalThirdCalculation.swift
+++ b/Rectangle/WindowCalculation/BottomVerticalThirdCalculation.swift
@@ -19,12 +19,20 @@ class BottomVerticalThirdCalculation: WindowCalculation {
         
         var calculation: WindowCalculation?
         
-        if last.action == .topVerticalThird || last.action == .bottomVerticalThird {
+        if last.action == .bottomVerticalThird {
             switch lastSubAction {
-            case .topThird:
-                calculation = WindowCalculationFactory.middleVerticalThirdCalculation
-            case .bottomThird:
+            case .centerVerticalThird:
                 calculation = WindowCalculationFactory.topVerticalThirdCalculation
+            case .bottomThird:
+                calculation = WindowCalculationFactory.middleVerticalThirdCalculation
+            default:
+                break
+            }
+        }
+        else if last.action == .topVerticalThird {
+            switch lastSubAction {
+            case .bottomThird:
+                calculation = WindowCalculationFactory.middleVerticalThirdCalculation
             default:
                 break
             }
--- a/Rectangle/WindowCalculation/TopVerticalThirdCalculation.swift
+++ b/Rectangle/WindowCalculation/TopVerticalThirdCalculation.swift
@@ -19,7 +19,7 @@ class TopVerticalThirdCalculation: WindowCalculation {
         
         var calculation: WindowCalculation?
         
-        if last.action == .topVerticalThird || last.action == .bottomVerticalThird {
+        if last.action == .topVerticalThird {
             switch lastSubAction {
             case .topThird:
                 calculation = WindowCalculationFactory.middleVerticalThirdCalculation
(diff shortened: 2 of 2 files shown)


khaledosama999:
@rxhanson  Let me know what you think, also I left a question about image creation for settings

rxhanson:
Thanks for the pull request! 

Sorry for the ambiguity, but the desired location for these additional shortcuts is here:
image

As for the icons, I have always hand drawn them in Pixelmator Pro, and they are actually quite simple to draw (or probably generate more accurately with further instruction) if you don't mind. They are template images (black only). The background rectangle is 40 percent opacity, and the frame and window rectangles are 100 percent opacity.

khaledosama999:
Ok makes sense my bad, will check it out today

khaledosama999:
Not the best UX in my opinion, I would've never have guessed we have extra shortcuts here (given it has its own tab and the arrow marker to display more shortcuts)

rxhanson:
> Not the best UX in my opinion, I would've never have guessed we have extra shortcuts here (given it has its own tab and the arrow marker to display more shortcuts)

Thanks for the feedback. I should have prefaced this with more context. The main feature request for Rectangle over the years is more sizes/positions. In order to accommodate all of these additional shortcuts, Rectangle's UI would have to change. I've tested out various interfaces, and in Rectangle Pro there is a table view and there is a custom size/position builder. For Rectangle, my desire is to reduce complexity as much as possible to make it more approachable - hence the single pane showing shortcut recorders. For many years, this meant only allowing additional sizes/positions to be contributed as Terminal command configurations. More recently, I've decided to allow more shortcuts to be added to the UI in this place, so that there's a middle ground that's more usable since Terminal command configuration of keyboard shortcuts is tedious. The fact that you would have never guessed we have extra shortcuts here is by design - it's inevitably going to be odds and ends that are added by the community. With that said, I'm always open to change.

khaledosama999:
> > Not the best UX in my opinion, I would've never have guessed we have extra shortcuts here (given it has its own tab and the arrow marker to display more shortcuts)
> 
> Thanks for the feedback. I should have prefaced this with more context. The main feature request for Rectangle over the years is more sizes/positions. In order to accommodate all of these additional shortcuts, Rectangle's UI would have to change. I've tested out various interfaces, and in Rectangle Pro there is a table view and there is a custom size/position builder. For Rectangle, my desire is to reduce complexity as much as possible to make it more approachable - hence the single pane showing shortcut recorders. For many years, this meant only allowing additional sizes/positions to be contributed as Terminal command configurations. More recently, I've decided to allow more shortcuts to be added to the UI in this place, so that there's a middle ground that's more usable since Terminal command configuration of keyboard shortcuts is tedious. The fact that you would have never guessed we have extra shortcuts here is by design - it's inevitably going to be odds and ends that are added by the community. With that said, I'm always open to change.

Ok man thanks, makes sense, will ping you when I update the PR 💯

khaledosama999:
@rxhanson  Done, images are still bad (I'm a back-end engineer, so UI is my kryptonite) let me know if all is good

khaledosama999:
Any updates  ? 👂👁️👃👁️👂

rxhanson:
I’m in the middle of holiday travel, but will give this a look soon. I’ll add the icons if needed prior to merging.

rxhanson:
Repeating the shortcuts for the thirds actions here cycles between center and back. My expectation is that it would cycle across all of them, like how it's done for the existing vertical thirds.

khaledosama999:
@rxhanson  Done I also did the same for the middle (not sure if thats also what you intended)

rxhanson:
Thanks for the update. Almost there, just a couple small things to address:
* Cycling the top or bottom thirds will infinite loop and crash
* The existing vertical center third doesn't cycle, so I would expect this horizontal center third to also not cycle

khaledosama999:
@rxhanson ok I think we're good, (stupid vibe coding)

rxhanson:
Thanks! A few more items:
* Cycling starting with bottom third should cycle up, rather than down (again, to match existing thirds).
* The bundle identifier for the app should not be modified
* There are comments for "New Vertical Thirds" that have been added by Claude. Probably best to simply remove those as they will no longer be "New" after this is merged.

Normally at this point in a pull request I would make these changes myself and merge, but in the case of AI assisted/generated pull requests I prefer having this done by the author.

khaledosama999:
Done

rxhanson:
Thanks!

PR #1684, opened 2025-12-30
Add tileActiveApp action to tile windows of the active application

hi2gage:
## Summary
- Adds a new `tileActiveApp` window action that tiles only windows belonging to the currently active (foremost) app into a grid
- Complements the existing `cascadeActiveApp` action with a tiling equivalent
- Follows the same implementation pattern as existing multi-window actions

https://github.com/user-attachments/assets/bc41ea9d-62b1-4d73-aa96-5724a6c29aea


rxhanson:
Thanks!

PR #1693, opened 2026-01-16
feat: Introduce 'Center Three Fourths' window action 

raunaqness:
Added a 'Center Three Fourths' for Ultrawide Monitors


rxhanson:
Thanks for looking to contribute. Looks like your fork of Rectangle is outdated. Center three fourths was added a couple of releases ago.

PR #1736, opened 2026-04-03
Preserve relative position when moving to another display

benoittgt:
Previously, moving a window to the next/previous display (or a specific display) would always center it on the destination screen when no prior Rectangle action was recorded (e.g. manually positioned windows).

Fix: compute the window's proportional position on the source screen and apply it to the destination screen, clamping to stay within bounds. The `attemptMatchOnNextPrevDisplay` path and maximize special-case are unchanged.

Fixes #1723

- [x] Tested on my machine and it is working as expected

### DIFF
--- a/Rectangle/WindowCalculation/NextPrevDisplayCalculation.swift
+++ b/Rectangle/WindowCalculation/NextPrevDisplayCalculation.swift
@@ -24,28 +24,45 @@ class NextPrevDisplayCalculation: WindowCalculation {
         }
 
         if let screen = screen {
-            let rectParams = params.asRectParams(visibleFrame: screen.adjustedVisibleFrame(params.ignoreTodo))
-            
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks for contributing! Finally got a chance to test this out. I'm thinking that this is intended, but going from right half on my laptop display to my next display resulted in: 

Untitled

This result exposes ambiguity about my expectation of where the window would be anchored, and is unintuitive enough to me that my preference is to keep the existing functionality as the default behavior and have your change be something configured via Terminal command, similar to `Defaults.attemptMatchOnNextPrevDisplay` seen here.

benoittgt:
I think the issue is somewhere else and with a conflict with Tahoe, as I can move window without Rectangle running. So I there there is a conflict. Closing.

PR #1746, opened 2026-04-19
Add screen presets feature

Diastro:
## Summary

Adds a new Settings tab for saving and restoring named window-layout presets, each with its own keyboard shortcut.

- **Capture**: Create a preset with `+` to snapshot every visible window of every running regular app, or use `Add App...` to pick a single installed app and capture one window at a time via a searchable popover.
- **Restore**: Each preset binds to a MASShortcut and also has a `Restore Now` button. Windows are placed proportionally using relative frames (0-1 fractions of the capture-time display), so the layout scales across different-sized displays. Apps that are installed but not running are launched first; uninstalled apps are pruned automatically.
- **Multi-display**: The primary display is the system primary (screen at origin 0, 0), not `NSScreen.main`. Each captured window records a `screenIndex` so you can move it between primary and secondary displays via `Edit App...`.
- **Storage**: Presets persist to `UserDefaults` under `screenPresets` and are deliberately kept out of the main Config export/import.

image


rxhanson:
Thanks for contributing!

My preference at the moment is to have this new settings tab hidden by default, and shown either by setting a default via Terminal command or via a checkbox in the popover in the ellipsis button at the bottom of the General tab. One of my main goals with Rectangle has always been to keep it really simple and approachable out of the box, although I view it as a community-oriented project so I want to keep it flexible with nearly anything people want to add in.

I'll dive into testing and reviewing the code soon.

PR #1759, opened 2026-05-09
feat(shortcuts): add duplicate shortcut cycling

tmsdnl:
Related Discussion: #1758 

**Summary**
- Cycle between actions that share the same keyboard shortcut
- Restart the cycle if the window was moved or resized after the last Rectangle action
- Keep Rectangle shortcuts suspended while MASShortcut recording is active
- Reject Todo shortcut conflicts quietly without the MASShortcut "already-used" alert
- Add coverage for cycling, Todo shortcut validation, and recording state

**Verification**
- `xcodebuild test -project Rectangle.xcodeproj -scheme Rectangle -destination 'platform=macOS' -derivedDataPath /private/tmp/RectangleDerivedData`
- `xcodebuild build -project Rectangle.xcodeproj -scheme Rectangle -destination 'platform=macOS' -derivedDataPath /private/tmp/RectangleDerivedData`

**Disclaimer**
This implementation was Codex-assisted.

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -40,6 +40,8 @@
 		8B1D14E3A55936EBCCC1E807 /* MiddleRightTwelfthCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA84AB4DB9E7EEFCE50AA6D8 /* MiddleRightTwelfthCalculation.swift */; };
 		924919761FC491952752A147 /* TopCenterLeftTwelfthCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9698D1D9EE0180D7C4468304 /* TopCenterLeftTwelfthCalculation.swift */; };
 		944F25CD2CE5A144004B2FD2 /* PrefsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 944F25CC2CE5A144004B2FD2 /* PrefsViewController.swift */; };
+		944F25CF2CE5A144004B2FD2 /* ShortcutRecordingObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 944F25CE2CE5A144004B2FD2 /* ShortcutRecordingObserver.swift */; };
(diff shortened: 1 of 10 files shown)


rxhanson:
Thanks for contributing! I quickly glanced through the changes and will give this a more thorough run-through soon.

I didn't quite catch it - how is the ordering determined for which action gets executed first if there are multiple actions with the same keyboard shortcut?

This change sort of recenters the ShortcutManager around this functionality, and I'm thinking there might be a more elegant way to achieve this while keeping it a little more separate. When I dive into this a bit deeper, I think I'll have a more clear picture of how to do this.

tmsdnl:
I see what you're saying. This leaks shortcut-recording UI state into ShortcutManager. The cycling logic may belong there, but this pause/resume behavior needs a cleaner boundary.

Let me know your thoughts and preferred direction.

rxhanson:
Sorry for the delay. The reason for thinking down the path of reducing some of the responsibility of the ShortcutManager is that I am planning to swap out MASShortcut and would like to keep some of the MASShortcut interaction as isolated as possible. I'm actually finding it difficult to draw the line in ShortcutManager, though - do you have any thoughts on a good way to do this? It's nice to have the struct for keeping the cycling organization separate, but maybe there's a good way to pull more of the ShortcutManager class modifications out. Aside from that things are looking good.

rxhanson:
Let me know if you have any further thoughts on a good way to split out some of the responsibility of the ShortcutManager, and if that's something you have a desire to continue working on. If not, that's ok - I'll roll in your change and split it out in a follow-on effort. Thanks again for contributing!

tmsdnl:
Hi Ryan, I got a little busy too. I’ll try to respond later this week, probably over the weekend.

rxhanson:
I'll go ahead and roll this in. I might have a more clear path when I roll in the MASShortcut rewrite.

tmsdnl:
Thanks Ryan, and sorry I didn’t get back to this sooner. Appreciate you pulling it in. The rewrite sounds like the right time to clarify that boundary. Let me know if I can be helpful in any way.

PR #1762, opened 2026-05-17
Add overlap offset for grid position cycling

MyronKoch:
## Summary

- When a window is placed at a position already occupied by another window, apply a small offset (default 11pt) so the user can see there is a window underneath
- Add "Offset cycling position on overlap" checkbox in the Grid Positions section of the extra settings popover
- Configurable via hidden preferences: `cyclingOverlapOffsetSize` (default 11), `cyclingOverlapMaxCascade` (default 1, max 5)
- Overlap detection uses origin-point matching, so mixed-size windows at the same grid position (e.g. quarter + eighth) also trigger the offset
- Feature is off by default

Rectangle offset dark
Rectangle offset

## Motivation

On large displays (4K TVs, ultrawides) with fine-grained grid positions (ninths, twelfths, sixteenths), placing a window at a position occupied by another window completely hides the window underneath with zero visual indication. This is especially common on multi-monitor setups where many windows share grid positions across workflows.

## Implementation details

- `WindowManager.applyOverlapOffsetIfNeeded` - core overlap detection and offset logic, gated behind `positionCycles` to avoid unnecessary AX enumeration on non-grid actions
- Clamping uses AppKit coordinates consistently; overlap detection converts to AX coordinates for comparison against window frames
- Guards against zero/negative offset, caps cascade at 5
- Three new defaults added to `Defaults.array` for config export/import

## Test plan

- [x] 18 unit tests: positionCycles property, screenFlipped math, Defaults.array inclusion, offset clamping with primary and secondary monitor coordinates
- [x] Manual QA on 4-screen setup (MacBook + three 4K TVs): overlap offset, mixed sizes, checkbox toggle, cycling sequence, multi-monitor
- [x] 10+ days of daily use on multi-monitor 4K setup


rxhanson:
Looks good, I like it! I'm going to test it out a little bit more and if I don't hit any snags I'll go ahead and merge.

I've been thinking about adding a feature where you can cycle the windows stacked on top of each other, and this would fit nicely with that concept. Looking forward to the hover count badge idea if that one comes in a follow-on.

rxhanson:
Ah, after merging I did end up finding a bug, perhaps with your most recent change. Now simply executing a left or right half will add an offset even if there is no underlying left or right half window.

PR #1763, opened 2026-05-20
feat: add Combined Display Mode

dooyeoung:
## Summary

- Adds a **Combined Display Mode** setting that treats multiple monitors as a single virtual display
- Window actions (left/right half, center, maximize, corners, etc.) operate across all monitors at once
- Useful for portrait monitors placed side by side where centering across displays is desired

## Settings

A new checkbox is added to the Settings window under the display-related options, separated by a divider:

> ☑ Treat multiple monitors as one
> When using multiple monitors, treats them as a single display. Requires System Settings > Desktop & Dock > Displays have separate Spaces to be OFF.

Localized via `NSLocalizedString` with Korean translation in `Main.xcstrings`.

When enabled, the Next/Previous Display menu items are hidden.

## Implementation

- `Defaults.combinedDisplayMode` (`BoolDefault`) persists the setting
- `WindowCalculationParameters` carries an optional `combinedDisplayFrame: CGRect?`
- `WindowManager` computes the union of all screen visible frames and passes it through the calculation pipeline
- `CenterCalculation`, `CenterProminentlyCalculation`, and `LeftRightHalfCalculation` updated to respect the combined frame
- `BestEffortWindowMover` receives the combined frame as `visibleFrameOfDestinationScreen` to prevent single-screen clamping

## Test plan

- [ ] Enable Combined Display Mode in Settings
- [ ] Verify Left Half / Right Half fills the left/right monitor entirely
- [ ] Verify Center places window in the center of the combined display area
- [ ] Verify Maximize fills the combined display area
- [ ] Verify corners (Top Left, Top Right, etc.) work correctly
- [ ] Verify Next/Previous Display menu items are hidden when mode is on
- [ ] Disable Combined Display Mode and verify normal per-monitor behavior is restored

### DIFF
--- a/Rectangle/PrefsWindow/SettingsViewController.swift
+++ b/Rectangle/PrefsWindow/SettingsViewController.swift
@@ -954,37 +954,7 @@ class SettingsViewController: NSViewController {
         
         initializeCycleSizesView(animated: false)
 
-        if combinedDisplayModeCheckbox == nil,
-           let parentStack = doubleClickTitleBarCheckbox.superview as? NSStackView,
-           let insertIdx = parentStack.arrangedSubviews.firstIndex(of: doubleClickTitleBarCheckbox) {
-            let checkbox = NSButton(checkboxWithTitle: NSLocalizedString("Treat multiple monitors as one", tableName: "Main", value: "", comment: ""), target: self, action: #selector(toggleCombinedDisplayMode(_:)))
-            checkbox.state = Defaults.combinedDisplayMode.enabled ? .on : .off
-            // Match storyboard checkbox content priorities to prevent vertical compression
-            checkbox.setContentCompressionResistancePriority(.required, for: .vertical)
-            checkbox.setContentHuggingPriority(.defaultHigh, for: .vertical)
-
-            // Must be set before inserting: NSStackView queries intrinsicContentSize once on
-            // insertion, so preferredMaxLayoutWidth=0 would give zero height permanently.
-            let descLabel = NSTextField(wrappingLabelWithString: NSLocalizedString("When using multiple monitors, treats them as a single display. Requires System Settings > Desktop & Dock > Displays have separate Spaces to be OFF.", tableName: "Main", value: "", comment: ""))
-            descLabel.font = NSFont.systemFont(ofSize: NSFont.smallSystemFontSize)
-            descLabel.textColor = .secondaryLabelColor
-            descLabel.translatesAutoresizingMaskIntoConstraints = false
-            descLabel.preferredMaxLayoutWidth = 500
-            descLabel.setContentCompressionResistancePriority(.required, for: .vertical)
-            descLabel.setContentHuggingPriority(.defaultHigh, for: .vertical)
-
-            let separator = NSBox()
-            separator.boxType = .separator
-            separator.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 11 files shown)


rxhanson:
Thanks for contributing! I actually implemented this in Rectangle Pro and simply didn't have time yet to test it thoroughly, officially release it, and port it to Rectangle. When I implemented it, I had simply adjusted the NSScreen extension in ScreenDetection to have this:

```swift
    func adjustedVisibleFrame(_ ignoreTodo: Bool = false, _ ignoreStage: Bool = false) -> CGRect {
        var newFrame = visibleFrame
        
        if !NSScreen.screensHaveSeparateSpaces && Defaults.combinedDisplayMode.enabled {
            newFrame = NSScreen.screens.reduce(CGRect.null) { $0.union($1.visibleFrame) }
        }
...
```

That avoided touching the WindowManager and other specific calculation files. 

There might be a reason why Claude took your path instead, but off the top of my head I'm not sure what it would be. Would you mind swapping out your implementation here with this and seeing if it does what you expect?

dooyeoung:
> Thanks for contributing! I actually implemented this in Rectangle Pro and simply didn't have time yet to test it thoroughly, officially release it, and port it to Rectangle. When I implemented it, I had simply adjusted the NSScreen extension in ScreenDetection to have this:
> 
> ```swift
>     func adjustedVisibleFrame(_ ignoreTodo: Bool = false, _ ignoreStage: Bool = false) -> CGRect {
>         var newFrame = visibleFrame
>         
>         if !NSScreen.screensHaveSeparateSpaces && Defaults.combinedDisplayMode.enabled {
>             newFrame = NSScreen.screens.reduce(CGRect.null) { $0.union($1.visibleFrame) }
>         }
> ...
> ```
> 
> That avoided touching the WindowManager and other specific calculation files.
> 
> There might be a reason why Claude took your path instead, but off the top of my head I'm not sure what it would be. Would you mind swapping out your implementation here with this and seeing if it does what you expect?

@rxhanson 
Thanks for the context — helpful to see how you approached it in Rectangle Pro.

I'll be honest, I didn't fully understand the background behind the original implementation, which is probably why it ended up going a different direction.

Swapped it out and it works as expected. One small difference though: I used return instead of assigning to newFrame:

if !NSScreen.screensHaveSeparateSpaces && Defaults.combinedDisplayMode.enabled {
    return NSScreen.screens.reduce(CGRect.null) { $0.union($1.visibleFrame) }
}

If we assign and continue, the rest of the function — Stage Manager strip, Todo sidebar, screen edge gaps — all apply based on self (a single NSScreen) to what is now a multi-screen union. The Todo sidebar width is calculated from self.visibleFrame.width but gets subtracted from the full combined width, which seemed off.

Rectangle/PrefsWindow/SettingsViewController.swift
dooyeoung, line 1: preview image
Image

rxhanson:
Great! I'll test this out soon and will plan on merging if nothing else comes up.

rxhanson:
We are going to want to keep `newFrame = NSScreen.screens.reduce(CGRect.null) { $0.union($1.visibleFrame) }`, since the todo mode and stage manager functionality would break without taking that into account. At the very least, there is a calculation that needs to be fixed for handling todo mode percentage to make sure that it takes into account the combined screen width. I can look into it further, when I get a chance.

rxhanson:
Got it sorted out. Thanks!

PR #1766, opened 2026-05-25
Fix false positive overlap detection for half/third actions

MyronKoch:
## Summary

Fixes the post-merge bug from #1762 where executing a left or right half would add an offset even when there is no underlying window at that position.

## Root cause

The overlap detection used origin-only matching (within 4px tolerance). When snapping to left-half at origin (0, 0), any other window on screen with its origin near (0, 0) - such as a maximized window - triggered a false positive.

## Fix

Now requires that the other window fits within the candidate's dimensions: origins must match AND the other window must be the same size or smaller. This prevents a maximized window (1920x1080) from triggering the offset on a left-half (960x1080), while still detecting an eighth (480x540) behind a quarter (960x1080) at the same grid position.

## Test cases

- Two left-halves at same position - offset triggers (correct)
- Eighth behind quarter at same origin - offset triggers (correct)
- Maximized window behind left-half - NO offset (fixed)
- Right-half vs left-half - NO offset (correct, different origins)

### DIFF
--- a/Rectangle/WindowManager.swift
+++ b/Rectangle/WindowManager.swift
@@ -228,8 +228,11 @@ class WindowManager {
             let candidateAX = candidate.screenFlipped
             let hasOverlap = otherWindows.contains { element in
                 let otherFrame = element.frame
-                return abs(otherFrame.origin.x - candidateAX.origin.x) < tolerance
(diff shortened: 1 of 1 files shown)


rxhanson:
Thanks for the incredibly quick fix! Looks good.

PR #1767, opened 2026-05-26
Delete SECURITY_AUDIT.md

JakubAnderwald:
This was a one-off report to elaborate on a security fix PR I pushed. No need to keep it permanently in the repo.


rxhanson:
I was going to ask about that but was thinking it was intentional, as a note for future security audits. Thanks for the cleanup!

PR #1768, opened 2026-05-26
fix: restore snap after session has ended from log out/sleep

tjagithub:
- Add session change handling to SnappingManager
- Added session change handling to hopefully deal with the issue where Rectangle no longer snaps after idling and session ends
- Added tests for session lock, sleep, cycling


rxhanson:
Thanks for contributing! Were you able to confirm that this does indeed fix snapping in this scenario? Either way, it's worth a shot to see if it fixes it, so I'll go ahead and merge.

PR #1771, opened 2026-06-02
Fix window width not applied correctly when resizing across displays with dock on right

DinoStray:
## Problem

When the macOS dock is positioned on the **right side**, moving a window from a smaller display to a larger display using window management shortcuts (e.g., right half) could result in the target width not being applied. The window would only cover a portion of the intended half-screen area.

The issue was reproducible with "repeated command" modes when windows crossed display boundaries — the window would end up narrower than expected on the destination display.

## Root Cause

macOS enforces a constraint that prevents windows from extending beyond the main screen's frame.maxX via the Accessibility API. When targeting a right-half position on a large external display (e.g., an ultrawide monitor positioned to the right of a built-in Retina display), the calculated frame's maxX could exceed the main screen's frame.maxX. The Accessibility API would silently reject or clamp the size change, leaving the window at a width that fits within the main display's bounds (e.g., 744px instead of the intended 1720px).

### 1. AccessibilityElement.setFrame — safe-resize strategy

When the target frame's maxX exceeds the main screen's maxX, the method now uses an alternative sequence: move to a safe X position first where the size constraint is not triggered, resize to the target dimensions, then move to the final position. This works around the macOS constraint while preserving the existing size-position-size approach for all other cases.

### 2. WindowManager.execute — width-aware retry logic

The cross-display retry logic previously only checked for **height** mismatches when deciding whether to re-apply the window frame. It now also checks for **width** mismatches, ensuring both dimensions are correctly applied when moving windows between displays.

## Testing

Tested on macOS 26 with a dual-display setup:
- Built-in Retina display (1512×982)
- LG Ultrawide external display (3440×1440) positioned to the right, with the macOS dock on the right side

Reproduction steps:
1. Place dock on the right side of the screen
2. Open a window on the built-in display
3. Press right-half shortcut → window correctly fills right half of built-in display ✓
4. Move cursor to external display
5. Press right-half shortcut → window correctly fills right half of external display ✓
6. Toggle left-half / right-half on both displays and verify consistent behavior ✓

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -121,6 +121,10 @@ class AccessibilityElement {
     /// The Accessebility API only allows size & position adjustments individually.
     /// To handle moving to different displays, we have to adjust the size then the position, then the size again since macOS will enforce sizes that fit on the current display.
     /// When windows take a long time to adjust size & position, there is some visual stutter with doing each of these actions. The stutter can be slightly reduced by removing the initial size adjustment, which can make unsnap restore appear smoother.
+    /// If the target frame extends beyond the main screen's right edge, we first move the window
+    /// to a safe position (left side), resize it, then move it to the final position.
+    /// This works around a macOS constraint that can prevent windows from being resized
+    /// when their maxX would exceed the main screen's frame.maxX.
     func setFrame(_ frame: CGRect, adjustSizeFirst: Bool = true) {
         let appElement = applicationElement
         var enhancedUI: Bool? = nil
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks for contributing! 

I think there might still be something to iron out prior to merging, as I'm not able to reproduce the issue being fixed here. Would you mind providing a little more details on what might need to be done to reproduce it, or a video of it not working correctly so I can figure it out? I've tried with MBP display set as main, and external display set as main, going in both directions (even though the problem description is from smaller display to larger), and I wasn't able to see the issue. 

It's not entirely clear to me why the setFrame adjustment is based on the main screen only since a window could be placed on a display other than the main screen, so I expect that when I can reproduce the issue then I'll have a clear picture.

rxhanson:
I'm still interested in this if you're looking to move it forward and want to provide more details on how I can reproduce it and validate that the fix is good. Thanks!

rxhanson:
I'll close this one out for now since I haven't heard back, but I'm definitely still interested if this does indeed solve an issue that you were experiencing.

PR #1772, opened 2026-06-03
revamps to file-based configuration

philocalyst:
combats the defaults caching by replacing with a lightweight alternative.


rxhanson:
Thanks for contributing! So that I better understand, what is the scenario or specific usage that leads to defaults caching being an issue?

I appreciate seeing this path implemented, although I think that there's a better architecture that would bypass defaults and the associated caching: better config file support. Right now, you can export and import the config as JSON, and you can also place the JSON file in the Application Support directory (see the readme). However, it's lacking proper dynamic reading and writing of the config file. iTerm2 is a good example of this done right, and ideally Rectangle would handle it in a similar way (and many of the direct defaults usages in the app would have to be properly wired up to Rectangle's configuration management along the way). Would you be interested in walking down this path instead? 

A few general notes as I was looking though the changes here:
1. The Apple-intended interaction with defaults is through the available APIs so my initial thought is that needing to bypass the API is going down an unintended path, and this is something I typically avoid.
2. There's a number of style changes that are rolled in as part of this pull request, and my preference would be to keep those as a separate pull request to keep this change more clear. 
3. My preference is typically pretty strongly to avoid adding more external dependencies where possible. Although this dependency looks fine, it doesn't appear to be widely used and Rectangle would then be responsible for auditing the security of this dependency if it were updated.

rxhanson:
Just wanted to follow up and gauge interest in going the alternate route I had proposed. Thanks!

philocalyst:
I totally would! Let me put together an update! Glad you're interested, and thanks for laying out your thoughts :)

rxhanson:
Let me know if you have any questions when you get a chance to try out the proposed config file improvement. I'll close this one out, since I think it makes more sense to start fresh when looking down that path. Thanks again!

PR #1773, opened 2026-06-03
Adds a "Skip gap on top screen edge" checkbox in Settings

GVonB:
**Change Description:**
Currently, window gaps apply underneath the MacOS menu bar. On displays without a notch, this can create a larger perceived top margin relative to the other screen edges.

This PR adds an optional setting to exclude the top screen edge while preserving the gaps for the other sides and between windows.

Default behavior stays the same. 

**Original behavior, applying padding under menu bar:**
CleanShot 2026-06-03 at 13 45 53

**Behavior with top edge gap disabled:**
CleanShot 2026-06-03 at 13 47 44
CleanShot 2026-06-03 at 13 50 00

**Settings Menu:**
CleanShot 2026-06-03 at 13 59 31@2x

### DIFF
--- a/Rectangle/Base.lproj/Main.storyboard
+++ b/Rectangle/Base.lproj/Main.storyboard
@@ -2889,7 +2889,7 @@
                                             </stackView>
                                             <button verticalHuggingPriority="750" verticalCompressionResistancePriority="1000" translatesAutoresizingMaskIntoConstraints="NO" id="GTP-tg-Gap">
                                                 <rect key="frame" x="0.0" y="308" width="500" height="16"/>
-                                                <buttonCell key="cell" type="check" title="Skip gap on top screen edge" bezelStyle="regularSquare" imagePosition="left" inset="2" id="GTP-tg-Cel">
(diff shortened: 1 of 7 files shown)


rxhanson:
Thanks, I like it! 

* I was trying to think of a better label for the checkbox, maybe something like `Remove gap on top screen edge`
* What are your thoughts on tailoring this specifically to notchless displays? We could then make this something like `Remove top gap on notchless displays`. There is some existing logic in Rectangle for determining if the display has a notch or not, which could be leveraged. It's for specifying the gap on the display with a notch as opposed to removing it. I could see rolling in another change where a slider for this is also added.

GVonB:
> * I was trying to think of a better label for the checkbox, maybe something like `Remove gap on top screen edge`

I definitely thought the wording was strange as well. Originally I had `Apply gap under menu bar`, but that would imply the box should be checked by default to preserve current behavior. I will switch it over to `Remove gap on top screen edge`

Personally, I think keeping the option as a simple toggle rather than notch detection is a better option at least for now. I found that I still preferred the top-gap removal even on my notch display, so if `Remove top gap on notchless displays` were to be added, it might make sense to make that a separate toggle to leave the option open.

Of course, up to you and I can adjust the PR in whatever direction you prefer.

rxhanson:
I agree and I prefer the simplicity as well. Thanks again!

PR #1777, opened 2026-06-11
add feature override green button

f7urry:
image

Added a new setting that overrides the macOS green button behavior. When enabled, clicking a window's green button maximizes it to fill the screen instead of entering native macOS full screen. Holding any modifier key or using the button's menu restores the default macOS behavior.

Core manager responsible for intercepting and handling green button clicks.

- Registers a low-level `ActiveEventMonitor` for `leftMouseDown` and `leftMouseUp` events
- On `mouseDown`: detects clicks on the full-screen button (with no modifier keys held), captures the window element and button frame, and consumes the event to prevent native full screen
- On `mouseUp`: if the cursor is still within the button bounds, dispatches a **maximize** or **restore** action based on the window's last recorded Rectangle action
- Listens to `greenButtonOverride` and `configImported` notifications to start or stop the monitor dynamically

### `Rectangle/AccessibilityElement.swift`

- Added `isFullScreenButton` computed property that returns `true` when the element's subrole is `.fullScreenButton`

### `Rectangle/AppDelegate.swift`

- Declared a `greenButtonManager: GreenButtonManager` property
- Instantiated `GreenButtonManager()` during app startup alongside other managers

### `Rectangle/Defaults.swift`

- Added `greenButtonOverride` as a new `BoolDefault` with key `"greenButtonOverride"`
- Registered `greenButtonOverride` in the `array` of all defaults

### `Rectangle/PrefsWindow/SettingsViewController.swift`

- Added `greenButtonOverrideCheckbox` outlet
- Added `toggleGreenButtonOverride(_:)` action that saves the setting and posts the `greenButtonOverride` notification
- Added `initializeGreenButtonOverrideCheckbox()` that programmatically inserts the checkbox and a description label into the settings stack view after `doubleClickTitleBarCheckbox`
- Updated `initializeToggles()` to sync the checkbox state on load

### `Rectangle/Utilities/NotificationExtension.swift`

- Added `Notification.Name.greenButtonOverride` for broadcasting setting changes to `GreenButtonManager`

### `Rectangle/mul.lproj/Main.xcstrings`

- Added localization entries for:
  - Checkbox label: *"Click green button to fill screen instead of full screen"*
  - Description label: *"Clicking a window's green button maximizes it to fill the screen instead of entering macOS full screen. Hold any modifier key or use the button's menu for the default macOS behavior."*

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -151,6 +151,7 @@
 		AA0ACC2F2864A86D0025E376 /* StageUtil.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA0ACC2E2864A86D0025E376 /* StageUtil.swift */; };
 		AA3A9E8B29230A82004EB8E5 /* CFExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA3A9E8A29230A82004EB8E5 /* CFExtension.swift */; };
 		AA49DD1129B8C1B100690E13 /* TitleBarManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA49DD1029B8C1B100690E13 /* TitleBarManager.swift */; };
+		AA49DD2129B8C1B100690E13 /* GreenButtonManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA49DD2029B8C1B100690E13 /* GreenButtonManager.swift */; };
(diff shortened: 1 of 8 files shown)


rxhanson:
Thanks! 

Click event monitors can trip up other apps, so it might make sense to keep only one instance running at a time, although this might be something that I will address post-merge. 

I do think the checkbox label and the description could use some improvement (although this is a minor thing). Maximize is the nomenclature used elsewhere in Rectangle (rather than fill screen). And I typically always see the button referred to as the green stoplight button. Any thoughts/improvements here are welcome.

Also worth noting: I have a bit of rearranging of the settings that I'll be doing for an upcoming release, so this setting might get moved around when that time comes. 

I'll give this a little more testing, but at first look I don't see much getting in the way of merging.

PR #1778, opened 2026-06-15
Anchor clamped windows to the snap edge instead of leaving a gap

LoriTira:
## Summary

When a window can't be resized to fill its snap zone, Rectangle leaves it at the zone's *leading* corner, producing a gap on the screen-edge side. The clearest repro is **FaceTime snapped to a half (when in a vertical video call)**: it keeps a fixed video aspect ratio, so on a wide display it clamps its own width and ends up floating near screen-center instead of flush against the screen edge. This also happens with the System Settings window for example.

This makes a clamped window anchor to the snap zone's screen edge(s), and stay centered on any axis it can't fill. It's all behind the existing `moveFixedSizeToEdge` default, which is **off by default**, so default behavior is unchanged.

## Why it happens today

- FaceTime is *resizable* (`isResizable()` is `true`), so it goes through the **standard** mover chain and never reaches `FixedSizeWindowMover`. `StandardWindowMover` sets the zone position, the app clamps the size, and `BestEffortWindowMover` only rescues windows that overflow *off* the screen — an under-filled window keeps its leading-corner position, leaving the gap.
- The existing `moveFixedSizeToEdge` edge-anchoring in `FixedSizeWindowMover` was gated on `sharedEdges.isCorner`, so it only fired for quarter snaps, never halves (a half shares three screen edges, so `isCorner` is false).

## Changes

- **`ClampedWindowAligner`** — a pure, unit-tested helper that repositions a window per axis: if the zone touches exactly one screen edge on that axis, anchor to it; otherwise (spans the full axis, or floats inside) center. It subsumes the old `matchSharedEdges` / `centerWindowRect` and drops the `isCorner` gate.
- **`FixedSizeWindowMover`** now delegates to it, so half-snaps of fixed-size windows anchor correctly too. (Flag off → `.none` shared edges → centers, exactly as before.)
- **`EdgeAlignmentWindowMover`** added to the standard chain to re-anchor resizable-but-clamped windows like FaceTime after `StandardWindowMover` runs.
- Documented the (now broader) `moveFixedSizeToEdge` flag in `TerminalCommands.md`.

Edge detection and anchoring use the gap-applied rect, so existing gap settings are respected.

## Test plan

- [x] New unit tests for `ClampedWindowAligner` (right half, left half, full-height half, top-right quarter, interior, exact-fit): 6/6 pass.
- [x] Full app builds.
- [x] Manual, with `defaults write com.knollsoft.Rectangle moveFixedSizeToEdge -bool true`:
  - FaceTime snapped to the right/left half now sits flush against the screen edge, vertically centered (previously a gap).
  - Quarters snug into the correct corner.
  - Normal windows that fill a half, and all behavior with the flag off, are unchanged.

### DIFF
--- a/Rectangle/WindowManager.swift
+++ b/Rectangle/WindowManager.swift
@@ -11,6 +11,7 @@ class WindowManager {
     init() {
         standardWindowMoverChain = [
             StandardWindowMover(),
+            EdgeAlignmentWindowMover(),
             BestEffortWindowMover()
         ]
         
--- a/Rectangle/WindowMover/FixedSizeWindowMover.swift
+++ b/Rectangle/WindowMover/FixedSizeWindowMover.swift
@@ -2,59 +2,26 @@
 
 import Foundation
 
-/// Handle windows that are a fixed size, default to centering them in the proposed window area
+/// Handle windows that are a fixed size. With `moveFixedSizeToEdge` enabled, anchor them
+/// to the snap zone's screen edges; otherwise center them in the zone (legacy behavior).
 class FixedSizeWindowMover: WindowMover {
-    
+
     func moveWindow(toRect rect: CGRect, resultParameters: ResultParameters) {
         let windowElement = resultParameters.windowElement
         let currentWindowRect: CGRect = windowElement.frame
-        
-        let sharedEdges = resultParameters.calcResult.initialRect.screenFlipped.sharedEdges(withRect: resultParameters.visibleFrameOfScreen.screenFlipped)
-        
-        if Defaults.moveFixedSizeToEdge.userEnabled, sharedEdges.isCorner {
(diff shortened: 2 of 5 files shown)


rxhanson:
Thanks! This is great. There's one thing that I'm thinking about on it - that perhaps it would be nice to provide the granularity of configuration up front for whether or not this should apply to halves. We could change `moveFixedSizeToEdge` to an IntEnumDefault, where 0 is disabled, 1 is this behavior, and 2 is gated on corners. In my testing it was nice for some apps to be centered on left/right half. I'm ok merging this and rolling that in as well, if you're not interested in it.

rxhanson:
I'll go ahead and merge this one in and adjust the moveFixedSizeToEdge with more granularity in a follow-on change. Thanks again!

LoriTira:
Great! And thank you for dealing with the moveFixedSizeToEdge adjusting later!

PR #1782, opened 2026-06-25
Migrate to a new architecture for managing configuration!

philocalyst:
Implements a fairly robust two-way sync mechanism! Replaces the old default caching and allows people to have portable configuration they can work with outside of rectangle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,10 @@ There are buttons for importing and exporting the config as a JSON file in the s
 
 Upon launch, Rectangle will load a config file at `~/Library/Application Support/Rectangle/RectangleConfig.json` if it is present and will rename that file with a time/date stamp so that it isn't read on subsequent launches.
 
+## Live config file
+
+If you want to interface with rectangle with a configuration file,  enable **Sync settings to a config file** in the settings tab. When enabled, Rectangle reads and writes `RectangleConfig.json` in a folder you choose (defaulting to `~/Library/Application Support/Rectangle/`). Continously two-way syncs.
+
 ## Preferences Storage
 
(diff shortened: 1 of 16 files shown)


rxhanson:
Thanks for giving this another go! This might take me a little while to work my way through this, since it's a significant change (and I'm in the middle of moving). I haven't done any testing yet, but at first glance it looks like user defaults updates should still trigger preferences updates in Rectangle, in the interest of preserving user defaults terminal command manipulation. I think I might prefer the original defaults logic to be additionally preserved so that the complexity of handling configuration via file is not used if not desired, but that hopefully would be a straightforward modification to make with your introduction of the PreferencesStore (forgive me if I've overlooked something here). I'll give this some more thorough testing and review soon.

philocalyst:
Completely understand! I would say that with a workflow like this the defaults command is made irrelevant with tools like [jq](https://github.com/jqlang/jq) or a text editor -- so I would argue for deprecation. But ultimately, should be simple to add back.

rxhanson:
I am thinking that I'd like to take this in a slightly different direction where the read/write of the config file is handled more entirely by the existing `Defaults` class and keeping the abstraction contained there. Would you be alright with me closing this PR and taking some of the concepts and code to construct a version of this myself that's a little more in line with how it currently works?

philocalyst:
That would be fine! Consider co-authoring me 🥹

Let me know if I can help! Can't wait to see this live :)

philocalyst:
Closing because @rxhanson has got it:)

rxhanson:
Thanks! (I'll add you as a co-author when this goes in :)

Issue #1673, opened 2025-12-10
Unable to obtain window id


datazen123:
- [ ] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: Sequoia 15.7.1
Rectangle version: 0.92 (98)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 
Each instance that I try to move window using keyboard shortcut or using the menu (for example i try 'Left Half'), 
Rectangle Logging says: 
2025-12-10T09:20:28+08:00: Unable to obtain window id
2025-12-10T09:20:28+08:00: Unable to obtain window id
2025-12-10T09:20:28+08:00: Unable to obtain window id
(this shows if I attampted 3 window moves)

This bug happens when my MacBook Pro (M2) was sleeping for a few hours.
If I reboot, Rectangle works normally (the bug goes away and Rectagle works fine).

I've tried closing and reopenning Rectangle: bug remains.

Please advise. Thank you,

rxhanson:
Thanks for reporting. This log message implies that either macOS or the frontmost app is not vending the necessary window ID information to Rectangle via the Accessibility API. From this perspective, it's a bug in macOS that I can't reproduce and I don't have a clear path to fixing it off the top of my head. I'll think about this one a bit more and see if I can figure out a workaround.

articulatesnail:
macOS version: Tahoe 26.0.1
Rectangle version: 0.93 (99)
Macbook Pro M1 (2021)

Rectangle Logging:
```
2026-01-27T13:37:57+08:00: Unable to obtain window id
2026-01-27T13:37:57+08:00: Unable to obtain window id (on commands)
2026-01-27T13:38:02+08:00: Unable to obtain the accessibility element with the specified attribute at mouse location
2026-01-27T13:38:10+08:00: Unable to obtain the accessibility element with the specified attribute at mouse location (on clicks within window)
```

This issue happens occasionally, and resolves itself on reboot similar to datazen123's findings.

rxhanson:
Closing out old issues, this one might have been resolved by #1788, otherwise we're likely looking at an issue on the macOS side.

Issue #1325, opened 2024-01-22
MacOS Sonoma(14.2.1) Unable to obtain window id


abdlkrmdlgr:
When i wake up my device, rectangle not working properly. There are not working any shortcut and action. After restart my device, it's working again. I guess it's related with macOS version, before update to Sonoma it did not happened.

- [ ] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 14.2.1
Rectangle version: Version 0.75 (81)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 

```
2024-01-22T10:03:16+03:00: Unable to obtain window id
2024-01-22T10:03:17+03:00: Unable to obtain window id
2024-01-22T10:03:17+03:00: Unable to obtain window id
2024-01-22T10:03:19+03:00: Unable to obtain the accessibility element with the specified attribute at mouse location
```

kurzdev:
Experiencing the same issue with macOS 14.2.1 and Rectangle v0.75 Build 81

rxhanson:
You could try unregistering Rectangle from accessibility privileges in macOS and re-adding it. If you do this, you'll want to follow these steps:

1. Close Rectangle if it's running
2. In System Settings -> Privacy & Security -> Accessibility, first disable Rectangle, then remove it with the minus button. (it's important to do both of those steps in that order)
3. Restart your mac.
4. Launch Rectangle and enable settings for it as prompted.

Or, if you're handy with the Terminal you can just execute this terminal command:
`tccutil reset All com.knollsoft.Rectangle`

Sometimes the accessibility registration can just be flaky.

webskigosc:
Hey 👋🏻 I have the same problem, after wake up my device, rectangle not working but if I again switch the laptop to sleep mode and immediately wake it up it starts to work properly. This always happens after the device has been in sleep mode for about maybe 3h+ (for a long time). Of course, I checked the above solution, but unfortunately, after switching to sleep mode, the problem returns.

macOS Sonoma - version: 14.2.1 
Rectangle - version 0.75 (81)
Logs looks the same as above.

rxhanson:
@webskigosc thanks for reporting. Sometimes other apps that are using accessibility permissions can cause accessibility instability issues like this. You could try pruning your list if there's anything unnecessary in there.

Aside from that, you could also try this terminal command to change the way that Rectangle obtains the window under the cursor:

[...]

Issue #1244, opened 2023-08-18
Rectangle keys don't work after mac sleep


andrew-kay-oxb:
- [ ] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version:  Ventura 13.4.1
Rectangle version: 0.70 (76)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 
lots of these:
```
2023-08-18T10:40:51+01:00: Unable to obtain the accessibility element with the specified attribute at mouse location
2023-08-18T10:40:52+01:00: Unable to obtain window id
```
Rectangle used to work perfectly with Monterey (thank you -- brilliant tool.) I updated recently to Ventura, and problems began.  After mac has been asleep for a significant period and needs to be woken up with password/login again, Rectangle doesn't operate.  Keys don't work, and menu options do nothing.  Quit and restart rectangle doesn't help.  _Short_ sleeps (woken up with touch id) are ok, as is reboot.
Thanks

rxhanson:
Thanks for reporting. This means that macOS is not providing the necessary accessibility information to Rectangle. Unfortunately, this one appears to be a macOS issue. Here's a similar issue: https://github.com/rxhanson/Rectangle/issues/640. There's a workaround in that issue that might work for you? Not great, but better than a reboot every time.

Issue #640, opened 2021-11-16
(mac monterey 12.0.1) Rectangle unresponsive after afk induced logout


braxtonowens:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

Forgive me if I am just an idiot, but ever since the recent mac update I have noticed that after being afk for a long period of time mac will log out of my user account. Upon re-logging, box seems to re unresponsive to drag snapping. When attempting to force a snap via menu bar icon I hear the mac error clicking sound. The only fix seems to be a reboot. Again, sorry if i am an idiot and am missing something trivial.  

macOS version: Monterey 12.0.1
Rectangle version: 0.49 (54)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."):

rxhanson:
Does this happen every time?

janwaechter:
I can confirm that the same behavior is also occurring on my machine. Every time my MacBook Pro wakes-up to the login screen after going to sleep/being afk for a while. To be clear, the App is still running and available in the menu bar. 

According to the logs, there seems to be an issue with the window id (see below). Also, I can recreate the same issue with other window-managers as well (e.g. Magnet).

macOS version: Monterey 12.0.1
Rectangle version: 0.49 (54)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): `Unable to obtain window id`

rxhanson:
Unfortunately, this appears to be a macOS issue. I’ll think about adding a restart on wake option to Rectangle, but that’s kind of an annoying workaround.

braxtonowens:
It’s not super inconvenient, just thought if would be good to have this
documented. Thanks for the fantastic software!

alexarraiza:
I have the same problem, but I found that you can get it working without a reboot, you just have to lock the screen (command + control + Q), then push escape to turn off the screen, and then you can just wake the mac again and rectangle will work.

macOS version: Monterey 12.0.1
Rectangle version: 0.49 (54)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): Unable to obtain window id

mortenbm:
> I have the same problem, but I found that you can get it working without a reboot, you just have to lock the screen (command + control + Q), then push escape to turn off the screen, and then you can just wake the mac again and rectangle will work.
> 

This worked for me! Thank you!

elmanolito:
[...]

PR #1788, opened 2026-07-07
Execute window actions when the window id is unavailable

johncarmack1984:
The failure in #640 (also #1244, #1325, #1673): after waking to the login window or an idle logout, every action beeps and nothing moves until you lock and unlock the screen. Restarting Rectangle doesn't clear it. Rebooting does.

The logs in those threads show what macOS is actually withholding. "Unable to obtain window id" on commands means the focused window element was retrieved fine and both id sources then failed: `_AXUIElementGetWindow`, and the window list frame+pid fallback. The hit-test failure on drags means `CGWindowListCopyWindowInfo` and the system-wide element are dead too. Per-app accessibility keeps working the whole time, which also matches Magnet, BetterTouchTool, and HyperSwitch breaking simultaneously: the window server stops vending cross-app window info until a lock/unlock resets the session. I don't think Rectangle can clear that state. It can stop needing the window id to act. The id only feeds history bookkeeping (restore rects, cycling, todo and stage checks), but `execute()` requires it in the same guard as the window element, so a bookkeeping miss fails the whole action with a beep, and the drag path bails the same way.

With this change the id is optional. Actions run without it and skip the id-keyed bookkeeping: cycling restarts from the first step, no restore rect gets recorded. Restore alone still requires it and beeps, since it's pure history. The drag path tolerates a missing id, and window-under-cursor gets a last-resort fallback that hit-tests the frontmost app's own AX windows, since both existing lookups depend on window server info.

Verified by stubbing the two failing calls (`_AXUIElementGetWindow` erroring, `CGWindowListCopyWindowInfo` empty) and driving `rectangle://execute-action?name=left-half` at a real TextEdit window: on main nothing moves, on this branch it snaps, and with the stubs removed behavior is unchanged and the tests pass. There's a new test pinning that calculated geometry never depends on the id. The one thing I can't confirm on a healthy machine is the drag fallback under the real condition. If someone from #640 or #1673 can run this branch after the next occurrence, that settles it.

Multi-window actions and todo reflow still need the real window list, so they stay degraded in the broken state. Left alone on purpose.

### DIFF
--- a/Rectangle/AccessibilityElement.swift
+++ b/Rectangle/AccessibilityElement.swift
@@ -368,6 +368,25 @@ extension AccessibilityElement {
             }
             return windowElement
         }
+
+        // Last resort for when the window server isn't vending window info (#640):
+        // the frontmost app's own accessibility windows don't depend on it.
+        if let frontAppElement = getFrontApplicationElement(),
+           let windowElements = frontAppElement.windowElements {
+            let windowElement = windowElements
+                .map { (element: $0, frame: $0.frame) }
+                .filter { !$0.frame.isNull && $0.frame.contains(position) }
+                .min { $0.frame.width * $0.frame.height < $1.frame.width * $1.frame.height }?
+                .element
+            if let windowElement {
+                if Logger.logging, let pid = windowElement.pid {
+                    let appName = NSRunningApplication(processIdentifier: pid)?.localizedName ?? ""
+                    Logger.log("Window under cursor frontmost app fallback matched: \(appName)")
+                }
+                return windowElement
+            }
+        }
+
         Logger.log("Unable to obtain the accessibility element with the specified attribute at mouse location")
         return nil
     }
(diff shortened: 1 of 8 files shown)


rxhanson:
Thanks! Looks good to me. What are your thoughts on some form of temporary/alternate bookkeeping on windows without ids? Perhaps we could hash the window attributes from the accessibility API to come up with a "best guess" unique identifier and still maintain some notion of the existing bookkeeping. I don't mind merging this without it, but figured I'd see if it was something you'd be interested in exploring and potentially making a change for.

johncarmack1984:
Turns out to be pretty feasible! Pushed.

The wrinkle with hashing attributes: the obvious ones are exactly what churns. We mutate the frame ourselves, and titles change on their own (a browser retitles on every tab switch), so that hash dies mid-cycle. What holds is the element's own identity. Separate fetches of the same window come back CFEqual with the same CFHash, same across retrieval paths, distinct across sibling windows, and the hash tracks the element rather than its attributes, so our own moves don't churn it.

So the new commit derives the stand-in from CFHash of the window element, high bit set to keep it out of the real id space, and it only fires when both real sources fail. Cycling, restore, and unsnap history come back in the broken state. About 15 lines plus tests.

rxhanson:
Awesome. Looks good, I will test this out soon.

PR #1789, opened 2026-07-07
Align clamped windows to snap edges by default

### LINKED ISSUE #185: Constrained-width apps don't move flush to the sides of the screen
*macOS version:* 10.14.6
*Rectangle version:* 0.31 (36)

First off, thanks for Rectangle! I had Divvy for years but it never "clicked" with me, and Rectangle totally hasn't had that problem, I use it all the time.

Was a tough call as to whether this was a bug report or a feature request :-) 

I noticed an unexpected behavior with apps that have a maximum width. When I take, say, Tweetbot (max width of 628px, apparently) and tell Rectangle to put it in the right half of the screen (720px wide), it obviously can't -- but what it _does_ is anchors the window at the halfway point of the screen:

Screen Shot 2020-07-02 at 3 42 39 PM

What I'd _expect_ it to do is put the window flush with the right side of the screen at its max width:

Screen Shot 2020-07-02 at 3 42 59 PM

Looking at the logs (below), I guess that would mean that Rectangle would have to note the difference between the proposed and result sizing and offset the position by the difference. I suspect the same problem happens for max-height apps aligned to the bottom of the screen.

Not sure if this is feasible or not but I figured I'd ask! 

```
2020-07-02T15:37:19-04:00: AX sizing proposed: (720.0, 877.0), result: (628.0, 877.0)
2020-07-02T15:37:19-04:00: AX position proposed: (720.0, 23.0), result: (720.0, 23.0)
2020-07-02T15:37:19-04:00: AX sizing proposed: (720.0, 877.0), result: (628.0, 877.0)
2020-07-02T15:37:19-04:00: rightHalf | display: (0.0, 0.0, 1440.0, 877.0), calculatedRect: (720.0, 23.0, 720.0, 877.0), resultRect: (720.0, 23.0, 628.0, 877.0)
```

culpen90:
## Summary
- align constrained and fixed-size windows to snap edges and corners by default
- share moveFixedSizeToEdge enum handling between fixed-size and resizable-clamped windows
- document the enum values and cover edge, corner, and centered modes in tests

Fixes #185

## Testing
- xcodebuild test -project Rectangle.xcodeproj -scheme Rectangle -destination 'platform=macOS'

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -23,7 +23,7 @@ class Defaults {
     static let snapEdgeMarginRight = FloatDefault(key: "snapEdgeMarginRight", defaultValue: 5)
     static let centeredDirectionalMove = OptionalBoolDefault(key: "centeredDirectionalMove")
     static let resizeOnDirectionalMove = BoolDefault(key: "resizeOnDirectionalMove")
-    static let moveFixedSizeToEdge = IntEnumDefault<EdgeAlignment>(key: "moveFixedSizeToEdge", defaultValue: .centered)
+    static let moveFixedSizeToEdge = IntEnumDefault<EdgeAlignment>(key: "moveFixedSizeToEdge", defaultValue: .edgesAndCorners)
     static let ignoredSnapAreas = IntDefault(key: "ignoredSnapAreas")
     static let traverseSingleScreen = OptionalBoolDefault(key: "traverseSingleScreen")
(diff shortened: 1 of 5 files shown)


rxhanson:
Initially my preference was to not change the default behavior here, but after thinking about this some more I'm good with it. I'm going to wait to merge this until after other pull requests get merged, since it will be easier to fix potential merge conflicts here if they arise.

PR #1791, opened 2026-07-17
Add a Settings toggle to keep window size when moving a maximized window to another display

kmundnic:
## Summary

Moving a maximized window to the next/previous display currently re-maximizes it to fill the destination display. This is already toggleable via the hidden `autoMaximize` default, but there's no UI for it, so most users don't know it exists.

This PR exposes it as a **"Maximize window when moved to another display"** checkbox in Settings, **checked by default so existing behavior is unchanged**. Unchecking it keeps the window's size and centers it on the destination display — preferable when displays differ in size (e.g. a window maximized on a laptop screen shouldn't grow to fill a larger external monitor). This restores the behavior Spectacle had.

## Changes

- `SettingsViewController`: add the checkbox + description, mirroring the existing `greenButtonOverride` / `combinedDisplayMode` programmatic checkboxes (property, toggle action, initializer, config-import refresh).
- `Main.xcstrings`: register the two new strings.
- `TerminalCommands.md`: document the `autoMaximize` default.

## Testing

Built and ran locally on a dual-monitor setup:

- **Unchecked:** a window maximized on the smaller display keeps its size and centers when moved to the larger display.
- **Checked (default):** the window re-maximizes to fill the larger display, as before.

### DIFF
--- a/Rectangle/PrefsWindow/SettingsViewController.swift
+++ b/Rectangle/PrefsWindow/SettingsViewController.swift
@@ -48,6 +48,7 @@ class SettingsViewController: NSViewController {
     private var cooperativeCornerResizeCheckbox: NSButton?
     private var combinedDisplayModeCheckbox: NSButton?
     private var greenButtonOverrideCheckbox: NSButton?
+    private var autoMaximizeCheckbox: NSButton?
     
     @IBAction func toggleLaunchOnLogin(_ sender: NSButton) {
         let newSetting: Bool = sender.state == .on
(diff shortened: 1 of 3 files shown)


rxhanson:
Thanks for contributing! Although this was purposefully left out of the UI to avoid having too many toggles, my plan is to rework some of the UI to allow a lot more things to go in to a different location. I'm ok moving forward with merging this in when I get a chance to review it, with the notion that I will move it to the new location when I have it ready.

kmundnic:
Thanks! That makes sense. I've just been missing this feature since I moved from Spectacle to Rectangle. No worries about moving it to the new settings section whenever that's ready. I've marked this ready for review.

I've also tested it myself on a dual-monitor setup: with the toggle off, a window maximized on the smaller display keeps its size and centers on the larger one; with it on, it re-maximizes as before. I've attached an image of the UI with the new toggle.

Let me know if you have questions or concerns!

UI

PR #1792, opened 2026-07-18
Prevent Todo shortcuts from firing during shortcut recording

culpen90:
## Summary

- Suspend Toggle Todo and Reflow Todo bindings while a shortcut field is recording.
- Prevent Todo settings changes from restoring those bindings during recording.
- Restore only bindings allowed by the current Todo and Todo Mode state.
- Add regression coverage for suspension and state-aware restoration.

## Context

Window-action bindings are already suspended during shortcut recording. Todo shortcuts use separate MASShortcut registrations, so they could remain active and trigger Rectangle behavior while a new shortcut was being entered.

## Testing

- Focused shortcut-recording suite: 6 tests passed.
- Full suite comparison: this branch and a clean checkout of current upstream `main` report the same 21 unrelated cooperative-resize assertion failures. The branch adds three passing tests, for 173 total tests versus 170 on the upstream baseline.

Related to #195

### DIFF
--- a/Rectangle/ShortcutManager.swift
+++ b/Rectangle/ShortcutManager.swift
@@ -193,6 +193,8 @@ class ShortcutManager {
                 shortcutIdentities = ShortcutCycle.shortcutIdentities(shortcutsByAction: currentShortcuts)
             }
         }
+
+        TodoManager.setShortcutBindingsSuspended(isRecording)
     }
 
     private func isRepeatAction(parameters: ExecutionParameters, windowElement: AccessibilityElement, windowId: CGWindowID) -> Bool {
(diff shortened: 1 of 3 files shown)


rxhanson:
Thanks for contributing! #195 is a little different from this scenario, and some of the keyboard shortcut recording behavior has changed since that issue was filed. I'll give this a thorough test, but it looks good up front. I'll probably remove the auto-close of #195 since it's still valid in a more general sense than just Todo mode.

culpen90:
Thanks for the context—that makes sense. I updated the PR description to reference #195 without auto-closing it, so the broader shortcut-recording issue stays open.

PR #1795, opened 2026-07-21
Add a hover badge and switcher for stacked windows

MyronKoch:
Follow-on to #1762 (grid-position overlap offset). On that PR you mentioned looking forward to the hover count badge idea as a follow-on — here it is.

## What it does

When the overlap offset stacks multiple windows at the same grid position, resting the cursor on the stack's top-left corner shows a small count badge and a list of the stacked windows — each with its app icon and title. Click a name to bring that window forward.

It turns overlap from something you can't see into a lightweight switcher for that grid slot. On a laptop or small screen where you can't spread windows out, a stack of tiles plus this becomes a de-facto tab bar for the spot.

## Opt-in and unobtrusive

- **Off by default**, behind a new `stackBadge` hidden default, with a checkbox ("Show stacked window badge on hover") in the General-tab additional-settings popover next to the overlap-offset toggle, and an entry in `TerminalCommands.md`.
- **Hover-only** — nothing on screen at rest.
- The badge is click-through and sits below the title-bar band, so the window's traffic lights stay clickable.

## How it works (stateless)

No window state is cached, so nothing can go stale across window moves, closes, Spaces, Stage Manager, or sleep/wake. A lightweight timer reads the cursor position; only a brief dwell on a grid corner (computed from screen geometry alone) triggers a single window-server query via `CGWindowListCopyWindowInfo`, which can't block on an unresponsive app. Accessibility is touched only to read the matched windows' titles, on a background queue with a messaging timeout.

## Testing

- Unit tests for the grid-corner geometry, including negative-origin (secondary-display) screens.
- Multi-model adversarial code review across correctness, concurrency, and view lifecycle.
- Manually verified in Light and Dark mode and across multiple displays; soak-tested over ~2 weeks of daily use on 2 separate Macs.

### Light mode
Screenshot 2026-07-22 at 8 23 25 AM

### Dark mode
Screenshot 2026-07-22 at 8 22 39 AM

### GIF

stack-badge-dark

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -92,6 +92,8 @@
 		9824704F22B189250037B409 /* BestEffortWindowMover.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9824704A22B189250037B409 /* BestEffortWindowMover.swift */; };
 		9824705122B28D7A0037B409 /* LeftRightHalfCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9824705022B28D7A0037B409 /* LeftRightHalfCalculation.swift */; };
 		983BBD6F253B609D000D223E /* FootprintWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 983BBD6E253B609D000D223E /* FootprintWindow.swift */; };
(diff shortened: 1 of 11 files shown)


rxhanson:
This looks awesome! Looking forward to testing this out!

rxhanson:
Sorry for the delay on getting this one in. It's great! I'm planning to move some of the settings around in the next release so that they are not quite so hidden.

PR #1796, opened 2026-07-21
Export Todo sidebar width units

offyotto:
When the Todo sidebar width unit is set to percent, exporting configuration and reimporting it silently reverts the unit to pixels, because Defaults.array includes todoSidebarWidth but not todoSidebarWidthUnit, so the unit is never written to or read from the config file.

This adds todoSidebarWidthUnit to Defaults.array immediately after todoSidebarWidth so the unit is exported and restored alongside the width. All other export behavior is unchanged.

Test: `xcodebuild test -project Rectangle.xcodeproj -scheme Rectangle -destination 'platform=macOS' -only-testing:RectangleTests/DefaultsExportTests` (testTodoSidebarWidthUnitInExportArray fails before the change and passes after).

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -152,6 +152,7 @@ class Defaults {
         todoMode,
         todoApplication,
         todoSidebarWidth,
+        todoSidebarWidthUnit,
         todoSidebarSide,
         snapModifiers,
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks!

PR #1802, opened 2026-07-31
Allow zero to disable minimum window size limits

### LINKED ISSUE #388: Window (Zoom) width can't go smaller, but using the mouse it can
- [x] _This is a new bug that hasn't been documented in a closed issue or in the Readme._
https://github.com/rxhanson/Rectangle/pull/125/ seems related, and implemented, but setting it to something small, even 0 or 0.01, does not seem to affect the issue described

When using the Smaller shortcut to resize the Zoom application, the window at some point stops resizing, but it can be resized even furthen when using the mouse. I don't know if this is an issue in general or if it only affects Zoom. 

The top window in the image shows the minimum size when resizing the window using the Smaller shortcut. The bottom window shows the minimum size when resizing the window using the mouse.

macOS version: Big Sur 11.2
Rectangle version: v0.43 (48)
`<url on its way>`

culpen90:
## Summary

- allow an explicit zero minimum-window fraction to disable Rectangle's screen-relative limit
- use a reusable `DoubleDefault` for minimum-window fractions while importing legacy float-backed configs
- preserve the existing 25% default when no preference is stored
- allow windows to reach an exact configured limit while retaining a one-point safety floor
- document the zero-value behavior

## Root cause

The minimum-size calculation treated zero as invalid and substituted the 25% default. On the reported 2560-point display, the next 30-point resize step from 665 to 635 points was rejected against a 640-point floor before any accessibility resize was attempted.

## User impact

Setting `minimumWindowWidth` or `minimumWindowHeight` to `0` now lets the application enforce its native minimum size. Existing installations without an explicit setting continue to use Rectangle's 25% default.

## Testing

- `ChangeSizeCalculationTests`: 7 passed
- all tests outside the two existing cooperative-resize test classes: 128 passed
- full scheme: 171 passed; the 10 failing cooperative-resize test cases reproduce unchanged on `upstream/main`
- `git diff --check`

Fixes #388

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -28,8 +28,8 @@ class Defaults {
     static let ignoredSnapAreas = IntDefault(key: "ignoredSnapAreas")
     static let traverseSingleScreen = OptionalBoolDefault(key: "traverseSingleScreen")
     static let useCursorScreenDetection = BoolDefault(key: "useCursorScreenDetection")
-    static let minimumWindowWidth = FloatDefault(key: "minimumWindowWidth")
-    static let minimumWindowHeight = FloatDefault(key: "minimumWindowHeight")
+    static let minimumWindowWidth = MinimumWindowFractionDefault(key: "minimumWindowWidth")
+    static let minimumWindowHeight = MinimumWindowFractionDefault(key: "minimumWindowHeight")
     static let sizeOffset = FloatDefault(key: "sizeOffset")
     static let widthStepSize = FloatDefault(key: "widthStepSize", defaultValue: 30)
     static let unsnapRestore = OptionalBoolDefault(key: "unsnapRestore")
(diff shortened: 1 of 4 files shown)


Rectangle/Defaults.swift
```diff
@@ -361,6 +362,33 @@ class FloatDefault: Default {
     }
 }
 
+class MinimumWindowFractionDefault: FloatDefault {
```
rxhanson: It might be nicer to have a more generic new Default class that doesn't subclass FloatDefault that is used for these values instead, as it could potentially be reused for similar scenarios in the future. These values were originally added prior to the defaultValue being part of the FloatDefault. It might also make sense to go with Double values instead of Float where possible, since Float is rarely used and confusingly CGFloat is actually a Double on modern macOS (If I would have known that up front we wouldn't see much FloatDefault usage).

If you're up for looking down that path, it's likely preferable, but I can see going with this simple subclass approach if that doesn't make sense.

culpen90: Thanks — updated this along the suggested path. The specialized `MinimumWindowFractionDefault` subclass is replaced with a reusable `DoubleDefault` that conforms directly to `Default`. The minimum width and height preferences now use `Double`, exported configs use a `double` field, and legacy `float` configs retain their previous zero/default import behavior, so the `-1` sentinel is no longer needed. Coverage now includes absent versus explicit zero, legacy float migration, and Double precision; `ChangeSizeCalculationTests` pass 7/7 and CI is green.

rxhanson:
Excellent, thanks!

PR #1804, opened 2026-08-07
Make snap area config export deterministic

### LINKED ISSUE #1468: Config export ordering is non-deterministic

macOS version: Sequoia 15.0
Rectangle version: v0.83 (89)

When exporting Rectangle's config, the order of the values in `defaults.landscapeSnapAreas.string` seems to be non-deterministic.

This is a bit confusing if you diff the file. Is it possible to make the order of the output consistent?

arhxam:
## Summary

- serialize snap-area dictionaries in a stable `Directional.rawValue` order
- preserve Rectangle's existing alternating key/value JSON schema
- canonicalize previously persisted valid snap-area JSON during config export
- preserve raw export behavior for unset, malformed, and unrelated JSON defaults

Fixes #1468.

## Verification

- `JSONDefaultSerializationTests`: 5/5 passing
- full `RectangleTests`: 210 tests run; the same 22 pre-existing assertions fail as on `main`, confined to `ActiveSideSplitRatiosCooperativeTests`, `CooperativeCornerResizeTests`, and `HalfSplitCornerCalculationTests`
- `git diff --check origin/main...HEAD`

## AI assistance

AI tooling assisted with investigation, implementation, and test generation. The final diff was reviewed against the issue, independently code-reviewed twice, and verified locally.

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -436,6 +436,18 @@ class IntDefault: Default {
     }
 }
 
+private struct SortedSnapAreaConfigs: Encodable {
+    let snapAreas: [Directional: SnapAreaConfig]
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        for (directional, snapAreaConfig) in snapAreas.sorted(by: { $0.key.rawValue < $1.key.rawValue }) {
+            try container.encode(directional)
+            try container.encode(snapAreaConfig)
+        }
+    }
+}
+
 class JSONDefault<T: Codable>: StringDefault {
     
     private var typeInitialized = false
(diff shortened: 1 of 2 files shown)


rxhanson:
Thanks for contributing, sorry for the delay on reviewing this.

Typically to guarantee sorted keys when encoding JSON, we can add `.sortedKeys` to the encoder in `JSONDefault`, something like this:

```
encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
```

I'm guessing that codex was looking to impact the least amount of defaults in Rectangle (unless there's something I'm missing). If that's the case, I'd prefer just adding this `.sortedKeys` line to sort all of the JSONDefaults when they're saved and not having a special cased struct in place.

rxhanson:
I changed the JSON default to encode with sorted keys. I'll close this one out, but feel free to reopen if that doesn't address your issue.

Issue #1645, opened 2025-09-18
`smallerHeight` doesn't work when the window height is max


mohammad-alisafaee:
- [x] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: Tahoe
Rectangle version: 0.90
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 

I followed these instructions to set shortcuts for `smallerHeight` and `largerHeight`: https://github.com/rxhanson/Rectangle/blob/main/TerminalCommands.md#make-smallermake-larger-height-only.

However, `smallerHeight` won't work when the height of the window is maximum (i.e. fills entire screen height). It works as expected otherwise.

To reproduce:
* Set shortcut for `smallerHeight`
* Restart Rectangle
* Set a window to maximum height
* `smallerHeight` won't work

rxhanson:
Thanks for reporting. I'll slate this one for the next release.

PR #1807, opened 2026-08-07
Allow smaller to shrink the height of a vertically maximized window

### LINKED ISSUE #1737: Smaller/Larger toggle only change width of the window after "Half" and "Center" Size Adjustments

macOS version: 26.4 (25E246)
Rectangle version: 0.95
Logs if applicable (In Rectangle menu, hold option, "View Logging..."):  N\A (for some reason no "View logging visible in rectangle Menu while holding option key)

YuriNachos:
Closes #1737.

## Problem

Under the combined `.smaller` command, a vertically-maximized window (Half / full-height) shrinks in width but not in height — it stays pinned to the top and bottom screen edges. `.smallerHeight` already had an exception for this (added in b97a353 / #1645), but `.smaller` was left asymmetric, so the combined shortcut cannot shrink a maxed-height window down in both dimensions.

## Changes

- `ChangeSizeCalculation`: the `againstTopAndBottomScreenEdges` exemption for `.smallerHeight` is extended to `.smaller` only when the new hidden preference is enabled, so **the default behavior is unchanged** — a full-height window keeps its height under Smaller unless the user opts in.
- New terminal command config `smallerShrinksMaximizedHeight` (off by default), enabled with:

```bash
defaults write com.knollsoft.Rectangle smallerShrinksMaximizedHeight -bool true
```

  It is declared as a `BoolDefault` and included in the exported config, and documented in `TerminalCommands.md` under "Make Smaller shrink the height of full-height windows". As with the other hidden preferences, it is read at launch, so the app needs a restart after setting it.

`.smaller` is conceptually `.smallerWidth` + `.smallerHeight`, so with the config enabled its height component honors the same edge exception `.smallerHeight` received in b97a353.

## Testing

- The three `ChangeSizeCalculationTests` assertions that encode the pinned-height behavior (`testExplicitZeroDisablesScreenFractionMinimum`, `testSmallerCanReachExactConfiguredMinimum`, `testSmallConfiguredScreenFractionAllowsIssueRegressionStep`) keep the original (default) expected rects.
- Added `testSmallerKeepsHeightOfFullHeightWindowByDefault`, asserting the unchanged default: a full-height window shrinks only in width.
- Added `testSmallerShrinksHeightOfFullHeightWindow`, a regression test for #1737 that enables `smallerShrinksMaximizedHeight` and asserts the window shrinks in both dimensions (height by `sizeOffset`, `y` by `floor(sizeOffset / 2)`).
- The `ChangeSizeCalculationTests` suite passes.

### Note on pre-existing host failures

Running the full `xcodebuild test` suite on this branch shows 21 failures in `ActiveSideSplitRatiosCooperativeTests` and `HalfSplitCornerCalculationTests`. These are pre-existing host-environment failures: the identical `xcodebuild test` run on a clean checkout of `main` produces the same 21 failures in the same suites. They are unrelated to `ChangeSizeCalculation` — the only app file this PR touches — and are expected to pass on CI's display configuration.

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -35,6 +35,7 @@ class Defaults {
     static let unsnapRestore = OptionalBoolDefault(key: "unsnapRestore")
     static let unsnapRestoreFromSizeChange = OptionalBoolDefault(key: "unsnapRestoreFromSizeChange")
     static let curtainChangeSize = OptionalBoolDefault(key: "curtainChangeSize")
+    static let smallerShrinksMaximizedHeight = BoolDefault(key: "smallerShrinksMaximizedHeight")
     static let relaunchOpensMenu = BoolDefault(key: "relaunchOpensMenu")
     static let obtainWindowOnClick = OptionalBoolDefault(key: "obtainWindowOnClick")
     static let screenEdgeGapTop = FloatDefault(key: "screenEdgeGapTop", defaultValue: 0)
@@ -134,6 +135,7 @@ class Defaults {
         widthStepSize,
         unsnapRestore,
         curtainChangeSize,
+        smallerShrinksMaximizedHeight,
         relaunchOpensMenu,
         obtainWindowOnClick,
         screenEdgeGapTop,
--- a/Rectangle/WindowCalculation/ChangeSizeCalculation.swift
+++ b/Rectangle/WindowCalculation/ChangeSizeCalculation.swift
@@ -7,6 +7,7 @@ class ChangeSizeCalculation: WindowCalculation, ChangeWindowDimensionCalculation
     let screenEdgeGapSize: CGFloat
     let sizeOffsetAbs: CGFloat
     let curtainChangeSize = Defaults.curtainChangeSize.enabled != false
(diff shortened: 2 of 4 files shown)


rxhanson:
Thanks for contributing! Rectangle has had this functionality for `Smaller` for a while, and changing it could feel like a breaking change for users that are used to it working this way. 

My preferred path forward for this is to keep it the way it currently works, but add a terminal command config that can be used to toggle on your desired behavior. Then, if there is demand to have it your desired way by default, then it's usually a sign that it's best as a checkbox in the UI, and it would fit well once I sort out the UI a little better to accommodate more toggles.

I don't mind adding in the Terminal command config if you don't have a desire to do it.

YuriNachos:
Thanks for the direction — that makes sense, so I went ahead and added the terminal command config myself. The new `smallerShrinksMaximizedHeight` default is off unless enabled, so the current behavior is completely unchanged out of the box:

```bash
defaults write com.knollsoft.Rectangle smallerShrinksMaximizedHeight -bool true
```

It's documented in TerminalCommands.md, and the tests now cover both the default behavior and the enabled mode. Happy to adjust the naming or hand it over to your own take if you'd prefer.

rxhanson:
Thanks!

PR #1813, opened 2026-08-11
Detect stacks of maximized windows in the overlap offset and badge

MyronKoch:
Stacking two maximized windows is the case where you most need to know something is buried — the top window covers the screen completely — but neither the overlap offset (#1762) nor the stacked-window badge (#1795) fires for it. Both quietly skip maximized windows.

**Two causes:**

1. **The offset never runs for maximize.** It's gated on `action.positionCycles`, and `.maximize` is in that list's exclusion set. `positionCycles` describes whether an action cycles through sizes on repeat — which maximize doesn't — but it's consumed in exactly one place: this gate. Maximize doesn't cycle, yet two maximized windows still land exactly on top of each other, so the offset is let through explicitly.

2. **Screen-covering windows are skipped by both features.** That exclusion is deliberate and load-bearing: a maximized window shares its origin with every half/corner placement, so matching one would offset all of them (#1766). But it only needs to hold while the window being placed is *smaller*. Two windows that both cover the screen and share an origin are a genuine stack:
   - Offset: the exclusion now lifts only when the window being placed also covers the screen.
   - Badge: tiled windows still win — screen-covering windows form the stack only when the tiled ones don't — so ordinary layouts count exactly as before.

Placing a half onto a maximized window still ignores it, so #1766 stays fixed.

Note the offset needs somewhere to shift into, which gaps provide (maximize applies gaps unless `applyGapsToMaximize` is disabled). With no gaps a maximized window is flush to the screen edges and the existing clamping leaves it where it is — no offset, no harm.

Four unit tests cover the badge's new stack selection, including that a lone maximized window can't inflate the count.

### DIFF
--- a/Rectangle/StackBadge/StackBadgeGeometry.swift
+++ b/Rectangle/StackBadge/StackBadgeGeometry.swift
@@ -70,6 +70,34 @@ enum StackBadgeGeometry {
         return best
     }
 
+    /// Indices of the cascade stack, when some candidates cover the screen.
+    ///
+    /// A window covering the screen shares its origin with every left/right
+    /// half and corner placement, so counting one would inflate the badge on
+    /// ordinary tiled stacks. Tiled windows therefore win: the stack is taken
+    /// from them whenever they form one. Only when they don't is the stack
+    /// taken from the screen-covering windows - a pile of maximized windows,
+    /// where the badge is the only way to see what's buried.
+    static func stackIndices(among origins: [CGPoint],
+                             coversScreen: [Bool],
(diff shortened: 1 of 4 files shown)


MyronKoch:
Folding this into #1810. The offset half (maximize eligibility, per-axis room) is there now, since it turned out to be the same code path as the scan hardening rather than a separate change. The badge half — a screen-covering window joining a stack the tiled windows form, without forming one on its own — goes with the badge work in #1808 instead, so each PR stays within one subsystem.

Closing to keep the queue honest.

PR #1822, opened 2026-08-21
Prevent Chromium Enhanced UI performance regression

### LINKED ISSUE #1751: it's tanking performance of chromium-based browser on JavaScript operation speedometer 3.1 benchmark
It was tanking my speedometer 3.1 benchmark from 58 to 42 and disabling it and restarting Chrome actually fix the issue

culpen90:
## Summary

- add automatic Enhanced UI handling for recognized Chromium browsers
- disable Enhanced UI when a recognized browser starts or activates, and leave it disabled after Rectangle window adjustments
- preserve restore behavior for other apps, VoiceOver, Switch Control, and existing explicit modes
- document the available modes and add regression coverage for the policy and preference migration

Fixes #1751

## Testing

- EnhancedUITests: 8 passed
- Full Rectangle test suite: 242 passed


rxhanson:
Thanks for tackling this one, too! I'm typically not thrilled about having to maintain a list of bundle identifiers, but I'm thinking that this might be the way this one has to go. Will test it out soon.

Issue #1674, opened 2025-12-13
Rectangle not respecting dock size


ryangrush:
I know this issue has been reported before, ex: https://github.com/rxhanson/Rectangle/issues/467, but I wanted to share the context of what I'm seeing. 

The key thing for me is this only became an issue when I switched to Mac OS Tahoe. As others have said it seems to only be an issue when I plug or unplug my MBP into my external display (via a KVM switch most of the time). I also can't fix it by just clicking "Maximize", for example, multiple times. I have to kill the Rectangle process and open the app again.

macOS version: Tahoe 26.1
Rectangle version: v0.92 (98)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."):

rxhanson:
Thanks for reporting. As you probably noticed from previous issues, Rectangle is passively asking for the screen dimensions, which makes it hard to tell why restarting Rectangle would fix it. This would suggest that there's some form of caching of screen dimensions macOS side that gets refreshed when an app with accessibility permissions is launched, and perhaps the only way to work around it would be to have Rectangle automatically restart itself when displays are changed. 

If you don't mind, when this breaks, can you send what the logs show when you execute a command, and when you restart it, a log of the correct behavior? I'd like to make sure I understand what Rectangle sees when it's broken.

ryangrush:
@rxhanson here we go. Again I basically just used my KVM switch to switch from one laptop to the other and back again.

```
2025-12-15T11:19:09-06:00: AX sizing proposed: (1361.0, 786.0), result: (1361.0, 786.0)
2025-12-15T11:19:09-06:00: AX position proposed: (76.0, 77.0), result: (76.0, 77.0)
2025-12-15T11:19:09-06:00: AX sizing proposed: (1361.0, 786.0), result: (1361.0, 786.0)
2025-12-15T11:19:09-06:00: almostMaximize, display: (0.0, 75.0, 1512.0, 873.0), calculatedRect: (76.0, 77.0, 1361.0, 786.0), resultRect: (76.0, 77.0, 1361.0, 786.0), srcScreen: Built-in Retina Display, destScreen: Built-in Retina Display, resultScreen: Built-in Retina Display
2025-12-15T11:19:10-06:00: AX sizing proposed: (1492.0, 853.0), result: (1436.0, 829.0)
2025-12-15T11:19:10-06:00: AX position proposed: (10.0, 44.0), result: (10.0, 44.0)
[...]

PR #1823, opened 2026-08-21
Correct stale Dock insets during window actions

### LINKED ISSUE #467: “Maximize” sometimes (but not always) fails to notice the Dock
The frustrating this about this is that I cannot repeat it reliably, but it has been happening enough for me to notice it, and for long enough that I finally decided to report it.

—-

**macOS version:** macOS version 11.4 (20F5065a) — this is the latest public beta of Mac OS 11, and I suppose it’s possible that this is beta-related.

**Rectangle version:**

/Applications/Rectangle.app:
	CFBundleShortVersionString: 0.46
	CFBundleVersion: 51

—-

I have a keyboard shortcut for `Maximize` that I use all of the time. What I have noticed happening “recently” (although I’m not sure exactly when) is that the window will be maximized and end up partially behind the dock (which is unhidden on the left side of the screen).

I know there has been at least 1 release of Rectangle since I first noticed this happening because I remember thinking “Oh, I wonder if this will fix the ‘maximize’ problem I’ve been having” but it did not.

When it happens, pressing the keyboard shortcut for `Maximize` will always fix it. And when I _try_ to make it happen intentionally, I never can, so I can’t seem to figure out what might be the common situation that triggers this behavior.

I also do toggle the Dock’s “auto-hide” setting on/off regularly, but I am fairly certain that what I am seeing is _not_ just the result of the dock being unhidden when an app window was already ‘Maximized’.

culpen90:
## Summary

- refresh Dock placement when AppKit reports a stale visible frame
- repair missing and phantom Dock reservations after display, edge, and auto-hide changes
- preserve menu bar and notch boundaries while failing closed on ambiguous accessibility geometry
- add coverage for Dock placements, display moves, auto-hide, invalid geometry, and combined displays

## Root cause

macOS can leave `NSScreen.visibleFrame` stale after Dock preference or display-topology changes. Rectangle used that frame directly, so the first Maximize or related action could overlap the Dock or retain space for its former location.

## Testing

- `DockVisibleFrameTests`: 16 passed
- full Rectangle test suite: 250 passed
- `git diff --check`

Fixes #467
Related to #1674

### DIFF
--- a/Rectangle/ScreenDetection.swift
+++ b/Rectangle/ScreenDetection.swift
@@ -155,13 +155,251 @@ enum ScreenOrdering: Int {
     case yThenMinX = 3
 }
 
+enum DockUtil {
+    // macOS can leave NSScreen.visibleFrame stale after Dock changes (#467).
+    // The Dock's AX list provides a live, fail-closed signal for its current screen and edge.
+    private static let bundleIdentifier = "com.apple.dock"
+    private static let edgeTolerance: CGFloat = 24
+    private static let minimumReliableInset: CGFloat = 8
+    private static let maximumInsetRatio: CGFloat = 0.33
+    private static var cachedSnapshot: Snapshot?
+    private static var snapshotInvalidationPending = false
+
+    private enum Edge: Equatable {
+        case left
+        case right
+        case bottom
+    }
+
+    struct Snapshot {
+        let screenFrames: [CGRect]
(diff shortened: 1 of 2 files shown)


rxhanson:
Nice! Thanks for giving this one a go. I'll test through it soon, but will be relieved to have this one solved.

Issue #144, opened 2020-05-05
Implement Microsoft Windows snapping behavior for easier snapping to corners


orena1:
I really like the tool, lately I got a to work on a windows pc, and I actually think that the way the snap is working on windows is better. It is hard to explain exactly the shortcuts, but it feel very natural.

https://www.zdnet.com/article/windows-10-tip-more-window-snapping-shortcuts-and-secrets/

Do you think it will be possible to implement a windows mode?

There is also a linux implementation - 
https://extensions.gnome.org/extension/1723/wintile-windows-10-window-tiling-for-gnome/

rxhanson:
I don't understand what you're asking for. Are you asking to have a preset that of shortcuts so that you don't have to manually configure them?

orena1:
Its not exactly that, the whole way snap  is handled in windows is different, see this video for example: 
https://www.youtube.com/watch?v=9RkfHNUhYiQ&t=182s

In windows you just click w+left or w+right and it moved the window relative to the location the window is, no need to remember any shortcuts.

As I said it is hard to explain, but if you ever next to a windows machine, try the snap shortcuts and you will see, its very intuitive.

rxhanson:
Ok, that video made it easier to understand what you're looking for. This makes sense to me, and I think it would be worthwhile to have.

However, I don't currently have the time to add it in myself. If anyone wants to work on it, I would definitely consider a pull request for it.

Windows is a little bit different in that the shortcuts are not configurable by the user (last I used Windows), so there's some edge case scenarios where this could behave weirdly for a user that doesn't have the default configuration in Rectangle. An implementation of this would most likely have to take into account the actual keycombo that results in each action, so that you can change for example what happens when a user presses the down keycombo if the window has been previously snapped to a side half. Knowing what keycombo was executed is not something that's tracked currently - just the resulting actions are kept track of. There's a decent amount to consider with this, so I don't see it as a trivial change.

orena1:
Thank you for taking the time to understand my request. I understand now that it is not easy to implement, but at least the request is out-there .

VariantXYZ:
As a suggestion, maybe something like the following could work as an interim watered down implementation?

* Map "Modifier + Direction" to the already existing 'Move Left', 'Move Right' modifiers.
* Mod + Up:
** If window height is not 'maximum': Maximize height
** If window height is 'maximum': Maximize width
* Mod + Down:
** If window height is not 'maximum': Minimize window
[...]

Issue #1820, opened 2026-08-19
Cycling sizes issue when it should not (checks width only not height)


picklejw:
- [ ] This is a new bug that hasn't been documented in a closed issue or in the Readme.

I do not see this bug reported, I have my config set for 'Repeated Commands' with cycle on side actions and cycle quadrants and size on side.

When I position window to lower right repositioning to upper right works correct and not cycle sizes. But if I am on lower right then keyboard to full right side instead of just filling up to top of screen it will do that but also 'cycle' to a different width (then 50% which is what I expect). Guessing some check when resizing that it is not checking previous height with width when resizing so it thinks it needs to cycle different widths.

macOS version: Tahoe 26.6.1
Rectangle version: 0.98
Logs if applicable (In Rectangle menu, hold option, "View Logging..."): 
2026-08-19T14:59:52-04:00: AX position proposed: (1504.0, 861.0), result: (1504.0, 861.0)
2026-08-19T14:59:52-04:00: AX sizing proposed: (1504.0, 831.0), result: (1504.0, 831.0)
2026-08-19T14:59:52-04:00: bottomRight, display: (0.0, 0.0, 3008.0, 1662.0), calculatedRect: (1504.0, 861.0, 1504.0, 831.0), resultRect: (1504.0, 861.0, 1504.0, 831.0), srcScreen: LG Ultra HD, destScreen: LG Ultra HD, resultScreen: LG Ultra HD
2026-08-19T14:59:55-04:00: AX sizing proposed: (2005.0, 1662.0), result: (1504.0, 831.0)
2026-08-19T14:59:55-04:00: AX position proposed: (1003.0, 30.0), result: (1003.0, 30.0)
2026-08-19T14:59:55-04:00: AX sizing proposed: (2005.0, 1662.0), result: (2005.0, 1662.0)
2026-08-19T14:59:55-04:00: rightHalf, display: (0.0, 0.0, 3008.0, 1662.0), calculatedRect: (1003.0, 30.0, 2005.0, 1662.0), resultRect: (1003.0, 30.0, 2005.0, 1662.0), srcScreen: LG Ultra HD, destScreen: LG Ultra HD, resultScreen: LG Ultra HD

Actions for log:
Positioned: Lower Right -> Positioned: Right (which cycles sizes as well as 'fills to top')

rxhanson:
Sorry for the delay, thanks for reporting. I was able to reproduce it and will slate a fix for this coming release.

[...]

PR #1824, opened 2026-08-21
Let half actions preserve the window's size on the other axis (opt-in)

X-Wei:
## Summary

Adds an opt-in mode in which Left Half, Right Half, Top Half and Bottom Half only change the axis they belong to (Left/Right → width, Top/Bottom → height) and keep the window's size and position on the other axis — the way Win + arrow keys tile windows on Windows, or keyboard tiling works on KDE.

With it, the four half shortcuts are enough to reach every half and quarter: Left Half then Top Half lands in the top left quarter, Bottom Half from there expands back to Left Half, and Right Half then Left Half fills the screen.

Off by default (`halvesPreserveOtherAxisSize`); nothing changes unless it is enabled.

Related: discussions #665 and #811 (where this was asked for, and you said you would be fine merging a good PR for it configurable via Terminal command), and #144. This is the narrow version discussed there: no separate "Windows mode" and no preset or extra shortcuts — the existing half actions simply take the window's current position into account, the logic lives in one self-contained class, and everything sits behind a single default.

## Behavior when enabled

Each half action looks at where the window currently sits along its own axis and leaves the other axis exactly as it is:

| Window along the action's axis | Left Half / Top Half | Right Half / Bottom Half |
|---|---|---|
| spans the whole axis (or has a size no half action produces) | dock to the left / top edge | dock to the right / bottom edge |
| docked to the left / top edge | inside a quarter: cycle sizes along this axis, keep the other (if the repeated-command setting resizes; otherwise no change); plain half: repeats as today (cycle sizes, move across displays, …) | expand to the whole axis |
| docked to the right / bottom edge | expand to the whole axis | inside a quarter: cycle sizes along this axis, keep the other; plain half: repeats as today |

Some sequences:

| Sequence | Result |
|---|---|
| Left Half, Top Half (or Top Half, Left Half) | top left quarter |
| top left quarter, Bottom Half | Left Half |
| top left quarter, Right Half | Top Half |
| top left quarter, Left Half | ⅔ wide top left quarter, then ⅓, then ½ again (height unchanged; with the default cycle sizes) |
| top left quarter, Top Half | ⅔ high top left quarter, then ⅓, then ½ again (width unchanged) |
| Right Half, Left Half | full width and height (same as Maximize, including its gap setting) |
| Left Half, Left Half | as today: cycle 1/2 → 2/3 → 1/3, move across displays, or nothing, depending on the repeated-command setting |
| untiled (floating, centered, …) window, any half | as today |

Untiled windows and repeated executions of a plain half deliberately fall through to the existing code, so the feature never changes what the repeated-command setting does. Inside a quarter the cycle is driven by the window's current size (the size after the one it has now, or the first selected size if it has none of them), not by the execution count, because the history there holds the corner action.

## How it works

- `HalvesPreserveOtherAxisSize.rect(for:)` recognizes the window's state on each axis purely from its current frame: an axis counts as "docked" when the window's origin and extent match (within 10 pt, for apps that resize in character cells) an edge-docked rect that a half action produces on that screen — at the active split ratio or any cycle size, with or without gaps applied. Anything else counts as "whole axis". It returns `nil` whenever the existing logic should run.
- `LeftRightHalfCalculation`, `TopHalfCalculation` and `BottomHalfCalculation` consult it first when the default is enabled; otherwise they are untouched.
- The result reports the geometrically equivalent action (`.topLeft` + `.topLeftQuarter`, `.leftHalf`, `.maximize`, …) as `resultingAction` / `subAction`, so gaps and the window history behave exactly as if that action had been executed. For that, `WindowCalculation.calculate` now forwards `RectResult.resultingAction` instead of ignoring it — every subclass that sets it already overrides `calculate`, so this is not a behavior change by itself.
- Drag to snap goes through the same calculations, so the footprint preview matches the snap. With the default unsnap-restore the dragged window is back at its original size when it snaps, so snapping behaves as before.

## Testing

- `HalvesPreserveOtherAxisSizeTests` (38 cases): every state transition, cycling inside quarters (both axes, custom cycle-size selection, non-50 % split ratio, gaps, the non-resizing repeated-command modes, no selected sizes), fallback to the existing logic for untiled windows and repeated plain halves, cycled 2/3 and 1/3 columns/rows, non-50 % split ratios, gaps (including the skip-top-edge option), matching tolerance, and the disabled path.
- Full suite on top of 64a9186 (v0.99): 272 tests, 0 failures.
- Used daily on macOS 26.6 since 2026-08-21 with 5 px gaps, 50 % splits and the default repeated-command setting (resize).

## Try it

A notarized Apple-silicon build of this branch (unofficial, signed with my own Developer ID) is attached to https://github.com/maxing-labs/Rectangle/releases/tag/v0.99-halvesPreserveOtherAxis-b109 — install, then tick the checkbox at the bottom of the Extras popover in Settings.

## Docs / UI

- `TerminalCommands.md`: new section "Make the half actions tile like Windows or KDE" with the `defaults write` command.
- The second commit moves the ordering of the selected cycle sizes from `RepeatedExecutionsCalculation.sortedCycleSizes()` into `CycleSize.sortedSelectedSizes()` so the new code can share it; the protocol extension just delegates to it.
- Settings gets a "Half actions preserve the window's size on the other axis" checkbox, unchecked by default, at the bottom of the Extras popover in the General tab (below the side split ratio rows, as requested), with a tooltip describing the Left Half → Top Half → Bottom Half round trip; the two strings are registered in `Main.xcstrings`. Happy to drop the checkbox and keep this terminal-only if you prefer.

### DIFF
--- a/Rectangle.xcodeproj/project.pbxproj
+++ b/Rectangle.xcodeproj/project.pbxproj
@@ -38,6 +38,7 @@
 		7BE578F32EFE11010083DAE3 /* ActiveSideSplitRatios.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7BE578F22EFE11010083DAE3 /* ActiveSideSplitRatios.swift */; };
 		7EC9E3BAD5BC57E49A7BCEE5 /* TwelfthsRepeated.swift in Sources */ = {isa = PBXBuildFile; fileRef = F3592F8E85CE82B8FA662A31 /* TwelfthsRepeated.swift */; };
 		866661F2257D248A00A9CD2D /* RepeatedExecutionsInThirdsCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 866661F1257D248A00A9CD2D /* RepeatedExecutionsInThirdsCalculation.swift */; };
+		7B4C0A1F2F1A000100C0FFEE /* HalvesPreserveOtherAxisSize.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7B4C0A1E2F1A000100C0FFEE /* HalvesPreserveOtherAxisSize.swift */; };
 		88763FEFFA3AECF8502604D6 /* LowerMiddleCenterLeftSixteenthCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 378C96D832D4CC6BCCF9D604 /* LowerMiddleCenterLeftSixteenthCalculation.swift */; };
 		895C73559E1ACEFBBE8DC32D /* UpperMiddleCenterRightSixteenthCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEBF3EA35F0C110A170EAD33 /* UpperMiddleCenterRightSixteenthCalculation.swift */; };
 		8B1D14E3A55936EBCCC1E807 /* MiddleRightTwelfthCalculation.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA84AB4DB9E7EEFCE50AA6D8 /* MiddleRightTwelfthCalculation.swift */; };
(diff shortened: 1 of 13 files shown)


rxhanson:
Thanks for contributing! Things look good at a quick glance. The list of checkboxes in the General tab is steadily growing and I haven't ironed out my preferred way to prevent it from turning into something unwieldy. Would you mind putting that checkbox at the bottom of the popover displayed from the "Extras"  button in the General tab for now? I think I'll end up with another tab for more settings instead of the Extras button, but I haven't decided where everything will go yet.

X-Wei:
Done — the checkbox now sits at the bottom of the Extras popover, below the side split ratio rows, and the General tab is untouched. Same title, tooltip and default (off); the commit is amended and force-pushed. Happy to move it again once you settle on a tab for these.

X-Wei:
Pushed a second commit (c5c209f) after using this for a couple of days: inside a quarter, the action for the edge the window is docked to now cycles sizes along its own axis instead of doing nothing — Left Half takes a top left quarter to ⅔ wide, then ⅓, then back to ½ (height unchanged), and Top Half does the same to the height. It follows the repeated-command setting (only the resizing modes cycle; the others still leave the quarter alone), and the current size is read from the frame the same way the docked state is, so no new state is tracked. Plain halves are unchanged and still go through the existing repeated-execution code.

The only touch outside the feature is that the ordering of the selected cycle sizes moved from `RepeatedExecutionsCalculation.sortedCycleSizes()` into `CycleSize.sortedSelectedSizes()` so both can share it (the protocol extension now delegates). Tests are at 38 for the feature, full suite 272 / 0 failures; the PR description and the "Try it" build (now build 109) are updated. Happy to squash the two commits if you prefer.

X-Wei:
One more thing, for after this is merged rather than for this PR: after using the opt-in mode for a while I found I also wanted to tile three windows in columns, so I have a small follow-up on top of these two commits — **when a window is docked at exactly two thirds, the action for the opposite edge shrinks it to the middle third instead of filling the screen** (Left Half, Left Half, Right Half → center column). I'll open it as a separate PR once this one lands, to keep this one as is. It's on my fork as [`halves-center-third`](https://github.com/maxing-labs/Rectangle/tree/halves-center-third) with a [test build](https://github.com/maxing-labs/Rectangle/releases/tag/v0.99-halvesPreserveOtherAxis-b111) if you're curious.

X-Wei:
Rebased onto v1.100 (no conflicts). b77e5ca touches the same repeated-execution path, but the two don't interact: with the default off nothing changes and the #1820 fix applies as is; with it on, the half actions only consult the new class first and still fall through to the existing code for plain halves. All 38 tests for this feature pass on top of v1.100, and the rest of the suite is unchanged — the 7 `HalfSplitCornerCalculationTests` cases that fail here fail identically on a clean v1.100 checkout (they exercise the cooperative cycling that b77e5ca now gates behind `cooperativeCornerResize`), so they are unrelated to this PR.

Nothing else pending on my side — happy to squash the two commits or adjust anything whenever you get to it.

rxhanson:
Sorry for the delay, thanks for rebasing! I'll be reviewing Rectangle PRs this weekend and expect to work through all of them.

PR #1838, opened 2026-09-02
Make Screen Edge Gap setting editable in UI

rutgerkraaijer:
I use the currently hidden `screenEdgeGapRight` and `screenEdgeGapsOnMainScreenOnly` settings to keep the notification area clear at all times. Since hidden settings aren't particularly user-friendly, this PR aims to make this feature more easily accessible and manageable.

What this PR changes:
- Added a `screenEdgeGapRightEnabled` setting, defaulting to false.
  - To have the setting hold a default value of 370 instead of 0 when not in use. Default 370 because that's what neatly keeps space for the notification area on the right.
- Made `screenEdgeGapsOnMainScreenOnly` setting default to true.
  - Since this setting isn't exposed currently, the main use case for the setting is the notification area, which only sticks around on the main display.
- Added a section to the "extras" settings to explain the feature, toggle `screenEdgeGapRightEnabled`, `screenEdgeGapsOnMainScreenOnly`, and set the value of `screenEdgeGapRight`.
- Added a scroll bar to the popout "extras" settings view, since the menu is now quite tall and won't fit on all displays.
  - Probably already a problem on certain display configurations, but with the extra section this is definitely needed.

I've intentionally limited this change to the right edge gap, since that has a clear use case. There is an argument that the other sides could be exposed in the same way as well of course.

### DIFF
--- a/Rectangle/Defaults.swift
+++ b/Rectangle/Defaults.swift
@@ -42,8 +42,9 @@ class Defaults {
     static let screenEdgeGapTop = FloatDefault(key: "screenEdgeGapTop", defaultValue: 0)
     static let screenEdgeGapBottom = FloatDefault(key: "screenEdgeGapBottom", defaultValue: 0)
     static let screenEdgeGapLeft = FloatDefault(key: "screenEdgeGapLeft", defaultValue: 0)
-    static let screenEdgeGapRight = FloatDefault(key: "screenEdgeGapRight", defaultValue: 0)
-    static let screenEdgeGapsOnMainScreenOnly = BoolDefault(key: "screenEdgeGapsOnMainScreenOnly")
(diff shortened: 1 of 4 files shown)


rxhanson:
Thanks for contributing!

* You can remove the informational text (people that want screen edge gaps already know why they want it)
* You can remove the first checkbox, since values > 0 mean it's being used.
* Change the input field label to `Right Edge Gap (px)`

That should make things more clear and reduce space.

rxhanson:
I was thinking about this a little more, and I think I'll go ahead and roll in the screen edge gaps settings for all edges, and put it directly on the General tab. I'll keep your "only on the main screen" setting, though. Apologies on declining this, but I appreciate you posting it and exposing a desired path!

rutgerkraaijer:
No worries about the decline, happy to have contributed in some way!

PR #1847, opened 2026-09-05
fix: explain when windows cannot fit a layout

### LINKED ISSUE #522: first two thirds not accurate?
if i have two windows open—left with 'first two thirds' and right with 'last third'—the left window will always overlap the right window by an inch or so. doesn't matter if it's two safari windows, or two completely different apps. 'left half' and 'right half' plus all other configurations work perfectly, however.

am i missing something super obvious? apologies if so.

culpen90:
When an app will not shrink a window enough for a snap layout, show a brief “Window size limited” message on the destination display after the final resize attempt. This explains overlaps such as a window with a 600-point minimum width placed in a 504-point third. The message dismisses automatically without taking focus.

Ignore rounding differences and position-only actions, and discard superseded cross-display retries so they cannot restore stale geometry or feedback. Document the app-imposed size limit and available layout workarounds.

fixes #522.

Validation:
- All 320 tests passed on macOS 26.6.2, including 10 new regression tests for size limits, rounding, invalid frames, and cross-display retries.
- Native warning rendering, repeated-show dismissal timing, and foreground app preservation checked.
- Safari-specific behavior was not manually tested.


rxhanson:
Thanks! A couple of minor feedback points:

* It'd be nice to have a defaults setting to allow this to be turned off, since there's the potential that it would be shown a lot for some users depending on their workflow and it could be undesirable.
* It could use a bit better wording. Maybe something like this:
**Minimum window size reached** `Unable to resize window smaller. Windows may overlap.`

rxhanson:
I went ahead and adjusted the wording and will roll in a default for this. Thanks again!

Issue #1465, opened 2024-09-17
Animate resizing via keyboard shortcuts?


Rylon:
Hi, I tried Sequoia's new window tiling and I still much prefer the way Rectangle works, especially the customisation, but the little animations in Sequoia when a window moves into the new location are very nice. Are those new APIs available for Rectangle to use at all? Would be the best of both worlds for me!

Thanks for a great app!

rxhanson:
Thanks for the feedback! Unfortunately, there is no way to do this as a 3rd party. I'm not aware of any new APIs for it.

Rylon:
Ah that's a shame, is it something you could do yourself, or would that be too laggy? Not to worry anyway, I stuck with Rectangle as the built-in approach was just too inflexible.

rxhanson:
Yes, unfortunately it would be laggy provided the options available. If anyone wants to give it a go and gets something good working, I would be fine merging a pull request for it.

Rylon:
Ok cool, no worries then, thanks again!

savked:
Just wanted to check whether anything has changed regarding this with the release of Tahoe 26?

rxhanson:
As far as I know, there are no new APIs that would allow a 3rd party to do this in a nice way.

PR #1848, opened 2026-09-06
Add default-off experimental window animations

kiteretsu903:
Adds **Animate windows (experimental)** to Snap Areas preferences, off by default. When enabled, same-display window actions transition over 300 ms, and dragging a snapped window away restores its size over 180 ms. Keyboard Restore also uses the transition when the saved frame is on the same display.

Related to #1465.

The implementation interpolates position and size using elapsed time and a critically damped curve with a longer settling tail, with up to 60 updates per second through Accessibility. Missed ticks are skipped. Repeated shortcuts retarget from the displayed frame while retaining the logical destination for cycling and restore history. Existing window movers perform the final correction and readback. Refused AX writes end interpolation and fall back to that correction path.

The existing Enhanced UI policy spans the transition and is restored on completion or cancellation. Reduce Motion, VoiceOver, and Switch Control bypass animation. A manual grab finishes an in-progress transition even when drag-to-snap is disabled. The preference is included in configuration import/export, with English, Simplified Chinese, Traditional Chinese, and Japanese labels.

Initial scope: cross-display window actions, fixed-size window handling, cooperative corner adjustments, and bulk tiling/cascading retain their existing behavior. Target applications can still impose minimum sizes or respond slowly, so this remains experimental.

Validation:

- Xcode 26.6 on macOS 27.0: Debug tests passed, 319 tests, zero failures.
- Release build passed.
- Added coverage for delayed ticks, cancellation and retargeting, failed writes, monotonic motion without overshoot, drag offsets, accessibility restoration, configuration export, and rapid-shortcut history preservation.
- Live settings check: default off, checkbox layout, and persistence after restart verified.
- Manual TextEdit check on macOS 27.0: snapping, keyboard Restore, and drag-to-restore confirmed smooth; the final curve was selected after comparison with an earlier timing curve.
- Broader app/framework and older-macOS visual coverage remains to be gathered; no universal smoothness claim is intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,12 @@ Drag a window to the edge of the screen. When the mouse cursor reaches the edge 
 | Bottom left, center, or right third                    | Respective third                       |
 | Bottom left or right third, then drag to bottom center | First or last two thirds, respectively |
 
+### Experimental window animations
+
+In the snap-area preferences, **Animate windows (experimental)** enables a short transition for window actions and restoring a snapped window's size, including dragging it away from an edge. It is off by default and is separate from **Animate footprint**, which only affects the snap preview.
+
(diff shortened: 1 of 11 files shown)


rxhanson:
I've been thinking about adding this for a while, and I'm interested in merging this. Is there a reason that you have closed this pull request?

PR #1849, opened 2026-09-07
Add optional snapping animations and a blurred snap preview

kiteretsu903:
This adds two options in Snap Areas, both off by default: **Animate windows (experimental)** and **Blur footprint**.

Snapping and restoring use smooth transitions, and the existing footprint animation follows the same easing curve. The blurred preview adapts to light and dark mode, with a subtle outline and outer shadow. Its tint uses the existing color and alpha preferences, defaulting to zero.

Animation remains experimental because repeatedly updating size and position may cause issues in some edge cases. Animations respect Reduce Motion, and blur respects Reduce Transparency.

This also balances the checkbox layout and updates the permission wording for macOS 27.


Rectangle/Defaults.swift
```diff
@@ -341,6 +345,29 @@ class StringDefault: Default {
     }
 }
 
+class FootprintAlphaDefault: Default {
```
rxhanson: Rather than making a new default subclass, you'll want to use the DoubleDefault class, which allows an explicit zero.

rxhanson:
Thanks! This is great :)

Looks like a merge conflict arose from another PR that I merged. Let me know if you need help resolving it.

Also, if you don't mind, can you clean up some of the comments that are added in here, as it looks like some of them are more for train of thought understanding than necessary communication in the codebase?

rxhanson:
On this note, I've given this a lot of thought prior to the PR, and one of the ideas I had for mitigating the limitations from the Accessibility API is to overlay a blurred window owned by Rectangle, move the actual window off the screen via Accessibility, resize the blurred window with native animation, then place the window at its destination as the animation concludes. This has its own set of usability issues, and doesn't work great for windows that don't fit the destination rect, but it might be worth exploring. Figured I'd mention it if you feel inclined to explore that route as well - no worries if not.

kiteretsu903:
@rxhanson Hi, Ryan, Thanks for quick response@
(1) fixed using DoubleDefault
(2) conflict resolved. Also improved the look of the "minimum window size reached" warning.
(3) improved comments
(4) For animation, I think your idea makes a lot of sense. I'll actually explore in that direction because I'm seeing some jumps on edge cases in my implementation. Will let you know once I have an update.
Thanks again for the helpful feedback!

rxhanson:
One edge case that's not working right is when `Repeated commands` is set to `cycle sizes on side actions`, then cycling on the right and bottom (like right half or bottom half) doesn't work. Even with the experimental disclaimer, this is something that would be quite nice to have working properly since it's really common to have that Repeated commands setting set. I don't mind taking a look when I get time if you don't get to it right away.

kiteretsu903:
> One edge case that's not working right is when `Repeated commands` is set to `cycle sizes on side actions`, then cycling on the right and bottom (like right half or bottom half) doesn't work. Even with the experimental disclaimer, this is something that would be quite nice to have working properly since it's really common to have that Repeated commands setting set. I don't mind taking a look when I get time if you don't get to it right away.

Yep, will look into this!

kiteretsu903:
@rxhanson  Hi Ryan, the right/bottom size cycling issue is fixed. I also tried your blurred-window approach, so there are now two options: Direct Resize (default) and a blur preview.

I’ve tested both across AppKit and browser windows comprehensively, including repeated commands, restore, size constraints, cross-display moves, and cancellation. Some edge cases like fast cross-display preview dragging still has some lag, so it remains experimental.

Thanks for reviewing it!

kiteretsu903:
@rxhanson This PR is finalized. Please take a look. Thanks!

rxhanson:
This is pretty awesome to see. Thanks for fixing the cycle resize issue and for implementing the blur preview approach. I'll give the updated code one more look through tomorrow, but the experience is good and don't think there are any additional feature changes that need to be made. 

I might play around with the blur preview a little bit after merging, as I'm wondering if there might be some adjustments that could make it a little distracting. It would technically be possible to determine the window type based on where the stoplight buttons are, so as to match the corner radius and "copy" having the correct stoplight buttons present in a matching location. It also might be better to use a solid color window background instead of blur, but I'm not sure. 

I'm excited to merge this change in.

Update: In using with the blur preview animation more, I think some acceleration would also help, on both moving the blur window and fading it out. Again, no pressure to make further changes, just some thoughts as I'm using it.

Also, looks like another merged PR caused a conflict, but I don't mind resolving it. I'll hold off on merging anything else until this one goes in.

kiteretsu903:
@rxhanson Just resolved the conflicts. Let's keep this version for now, and I'll keep an eye on the further optimization like acceleration as well!

rxhanson:
I've been continuing to test the blur preview animation out and hit a snag where windows that show the window size warning have a lot of lag with their positioning (thanks for making that warning look nicer, by the way). 

The lag is painful enough, and it's a common enough scenario that I was thinking through a couple of options: 
1. Keep working on it
2. Remove the option for the blur preview animation for now and solve the issue in a follow-on fix.

Since this PR is already pretty big, I'm leaning toward option 2, just so I can keep things moving forward, but am open to either path. 

Still working my way through the code.

kiteretsu903:
@rxhanson Hey, Ryan. Agree that we can split blur preview into another branch, removing from this PR, while keeping other changes.

I can go ahead to do the split. Thanks for testing!
(btw i have been a rectangle user for a few years since intel mac and am very happy to contribute)

kiteretsu903:
@rxhanson 
Further fixed some issues and added an independent animation curve path optimized for keyboard-triggered snapping. Thought we could finalize this one!
I also split the blur preview path into another PR https://github.com/rxhanson/Rectangle/pull/1859

rxhanson:
Thanks! I removed the markdown and the main.swift entry point, and added back in the @main entry point to the AppDelegate. I'm going to go back through the WindowManager changes with fresh eyes tomorrow morning and will merge then as long as I don't have anything I want to change there 👍

PR #1853, opened 2026-09-12
fix: don't realign a window that is within a point of filling its zone

### LINKED ISSUE #1852: Maximize leaves window at x=1 and 1pt narrow when Dock is on the right

macOS version: 15.7.9
Rectangle version: 1.100 (106)

```
2026-09-11T21:46:57+02:00: AX sizing proposed: (839.0, 1079.0), result: (839.0, 1079.0)
2026-09-11T21:46:57+02:00: AX position proposed: (0.0, 38.0), result: (0.0, 38.0)
2026-09-11T21:46:57+02:00: AX sizing proposed: (839.0, 1079.0), result: (839.0, 1079.0)
2026-09-11T21:46:57+02:00: leftHalf, display: (0.0, 0.0, 1679.0, 1079.0), calculatedRect: (0.0, 38.0, 839.0, 1079.0), resultRect: (0.0, 38.0, 839.0, 1079.0), srcScreen: Built-in Retina Display, destScreen: Built-in Retina Display, resultScreen: Built-in Retina Display
2026-09-11T21:47:23+02:00: AX sizing proposed: (1679.0, 1079.0), result: (1678.0, 1079.0)
2026-09-11T21:47:23+02:00: AX position proposed: (0.0, 38.0), result: (-0.0, 38.0)
2026-09-11T21:47:23+02:00: AX sizing proposed: (1679.0, 1079.0), result: (1678.0, 1079.0)
2026-09-11T21:47:23+02:00: AX sizing proposed: (1678.0, 1079.0), result: (1678.0, 1079.0)
2026-09-11T21:47:23+02:00: AX position proposed: (1.0, 38.0), result: (1.0, 38.0)
2026-09-11T21:47:23+02:00: AX sizing proposed: (1678.0, 1079.0), result: (1678.0, 1079.0)
2026-09-11T21:47:23+02:00: maximize, display: (0.0, 0.0, 1679.0, 1079.0), calculatedRect: (0.0, 38.0, 1679.0, 1079.0), resultRect: (1.0, 38.0, 1678.0, 1079.0), srcScreen: Built-in Retina Display, destScreen: Built-in Retina Display, resultScreen: Built-in Retina Display
```

The first three maximize lines are `StandardWindowMover`: the requested width of 1679 comes back as 1678. The next three are `EdgeAlignmentWindowMover` re-centering that 1678-wide window in the 1679-wide zone, which moves it to x=1.

## Description

With the Dock on the right, Maximize leaves the window 1 pt narrower than the screen and shifted 1 pt to the right (x=1 instead of x=0). Left Half lands at x=0 as expected. It happens with every app I tried (Brave, Zed, Finder).

## Steps to reproduce

1. Put the Dock on the right side of the screen.
2. Run Left Half on any window.
3. Run Maximize.
4. The window ends at x=1 with a width of visible width minus 1.

## Measurements

Built-in display 1728x1117, visible frame (0, 0, 1679, 1079), Dock on the right. Window bounds read from `CGWindowListCopyWindowInfo` after each action:

| start → action | requested (x, width) | result (x, width) |
|---|---|---|
| Left Half → Maximize | 0, 1679 | 1, 1678 |
| full width (set via Finder AppleScript) → Maximize | 0, 1679 | 0, 1679 |
| Left Half → Last Two Thirds | 560, 1119 | 560, 1119 |
| Right Half → Last Two Thirds | 560, 1119 | 561, 1118 |
| {x 840, width 500} → Right Half | 840, 839 | 841, 838 |
| {x 600, width 500} → Right Half | 840, 839 | 840, 839 |

## Analysis

Two things combine:

1. macOS clamps an AX size change that grows the window's right edge onto the Dock edge: requesting a width that reaches x=1679 yields 1 pt less. A position change that brings the window to the same edge is not clamped, which is why Left Half → Last Two Thirds (resized at x=0, then moved) is exact while Right Half → Last Two Thirds (resized at x=840) is not. Finder's own AppleScript `bounds` reaches 1679, so the app itself is not limiting the width.
2. `EdgeAlignmentWindowMover` / `ClampedWindowAligner` (#1778, enabled by default in #1789) treats that 1 pt shortfall as a clamped window. Maximize shares both the left and right screen edges, so the window is centered: `round((1679 - 1678) / 2) = 1`, giving x=1. For right-anchored actions the window is anchored to the right edge instead, so the 1 pt gap moves to the left side.

Before #1778 the same 1 pt shortfall stayed next to the Dock, where it was barely visible.

## Possible fixes

- Skip realignment in `ClampedWindowAligner` when the shortfall on an axis is 1 pt or less.
- Or avoid the clamp in `AccessibilityElement.setFrame`: when the target's right edge is the Dock edge, apply the size slightly to the left, then move the window into place, since position changes are not clamped.

Related, but a different trigger: #1825.

---

Investigation done with Claude for a real bug I'm having.

anandghegde:
Fixes #1852.

## Problem

With the Dock on the right, Maximize leaves the window 1pt narrower than the screen and shifted 1pt right (x=1 instead of x=0). From the log in the issue, on a 1679pt-wide visible frame:

```
AX sizing proposed: (1679.0, 1079.0), result: (1678.0, 1079.0)
AX position proposed: (1.0, 38.0), result: (1.0, 38.0)
maximize, calculatedRect: (0.0, 38.0, 1679.0, 1079.0), resultRect: (1.0, 38.0, 1678.0, 1079.0)
```

Two things combine, as diagnosed by @nyg in the issue:

1. macOS shortens an AX size change that grows a window's right edge onto the Dock edge, so a requested width of 1679 comes back as 1678. A position change to the same edge is not clamped, which is why Left Half → Last Two Thirds (resized at x=0, then moved) is exact while Right Half → Last Two Thirds is not.
2. `ClampedWindowAligner` reads that 1pt shortfall as a clamped window. Maximize shares both left and right screen edges, so the window gets centered: `round((1679 - 1678) / 2) = 1`.

The net effect is that the 1pt gap moves off the Dock edge, where it was invisible, onto the opposite edge, where it is not.

## Fix

Skip realignment on an axis when the shortfall there is within a point. I took the first of the two options suggested in the issue — the alternative, applying the size to the left of the target and then moving the window, would change `AccessibilityElement.setFrame` behavior for every action to address a 1pt cosmetic issue.

The check is per axis, so a window that is a point narrow but genuinely short still gets aligned vertically. Real clamps — a window holding a fixed aspect ratio, for example — are unaffected, since those miss the zone by far more than a point.

## Tests

Five cases added to `ClampedWindowAlignerTests`, using the measurements from the issue:

- `testMaximizeOnePointNarrowStaysPut` — the reported case (zone 1679 wide, window 1678, all four edges shared)
- `testRightHalfOnePointNarrowStaysPut` — the right-half variant from the issue's measurement table
- `testOnePointShortHeightStaysPut` — same tolerance on the vertical axis
- `testOnePointNarrowLeavesOtherAxisAlignmentIntact` — a point narrow *and* genuinely short: width is left alone, height still centers
- `testShortfallJustOverToleranceStillAligns` — pins the boundary so the tolerance can't be widened later without a test failing

The first four fail on `main` and pass with this change. The boundary test passes both ways by design. Full suite: 338 tests, 0 failures.

### DIFF
--- a/Rectangle/WindowMover/WindowMover.swift
+++ b/Rectangle/WindowMover/WindowMover.swift
@@ -14,10 +14,16 @@ protocol WindowMover {
 /// `.top` edge corresponds to maxY and `.bottom` to minY (matching the rest of Rectangle).
 enum ClampedWindowAligner {
 
+    /// A window may come back a hair smaller than its zone without being clamped in any
+    /// meaningful sense: macOS shortens an AX size change that grows a window onto the Dock
+    /// edge by a point. Realigning that shortfall moves the gap off the Dock edge, where it
+    /// was invisible, onto an opposite edge where it is not, so leave the axis alone instead.
+    static let alignmentTolerance: CGFloat = 1.0
+
     static func aligned(window: CGRect, inZone zone: CGRect, sharedEdges: Edge) -> CGRect {
         var result = window
 
-        if window.width != zone.width {
+        if abs(window.width - zone.width) > alignmentTolerance {
             if sharedEdges.contains(.left), !sharedEdges.contains(.right) {
                 result.origin.x = zone.minX
             } else if sharedEdges.contains(.right), !sharedEdges.contains(.left) {
@@ -27,7 +33,7 @@ enum ClampedWindowAligner {
             }
         }
 
-        if window.height != zone.height {
+        if abs(window.height - zone.height) > alignmentTolerance {
             if sharedEdges.contains(.top), !sharedEdges.contains(.bottom) {
                 result.origin.y = zone.maxY - window.height
             } else if sharedEdges.contains(.bottom), !sharedEdges.contains(.top) {
--- a/RectangleTests/RectangleTests.swift
+++ b/RectangleTests/RectangleTests.swift
@@ -4282,6 +4282,45 @@ class ClampedWindowAlignerTests: XCTestCase {
         XCTAssertTrue(result.equalTo(window))
     }
 
+    // A maximize against a Dock on the right comes back a point narrow (see #1852). Centering
+    // that shortfall would put the window at x=1 and leave a visible gap on the left edge.
+
+    func testMaximizeOnePointNarrowStaysPut() {
+        let zone = CGRect(x: 0, y: 0, width: 1679, height: 1079)
+        let window = CGRect(x: 0, y: 0, width: 1678, height: 1079)
+        let result = ClampedWindowAligner.aligned(window: window, inZone: zone, sharedEdges: [.left, .right, .top, .bottom])
+        XCTAssertTrue(result.equalTo(window))
+    }
+
+    func testRightHalfOnePointNarrowStaysPut() {
+        let zone = CGRect(x: 840, y: 0, width: 839, height: 1079)
+        let window = CGRect(x: 840, y: 0, width: 838, height: 1079)
+        let result = ClampedWindowAligner.aligned(window: window, inZone: zone, sharedEdges: [.right, .top, .bottom])
+        XCTAssertTrue(result.equalTo(window))
+    }
+
+    func testOnePointShortHeightStaysPut() {
+        let zone = CGRect(x: 1000, y: 0, width: 1000, height: 1200)
+        let window = CGRect(x: 1000, y: 0, width: 1000, height: 1199)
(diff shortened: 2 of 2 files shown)


rxhanson:
Thanks for the prompt fix!

Issue #1784, opened 2026-06-29
"Make-Smaller" or "Make-Larger" revert when dragging window


rrshaban:
I like this app.

I like it very much!

Thank you.

When I apply ctrl_option_- and ctrl_option_+, "Make-Smaller" and "Make-Larger", they decrement the size of the window by the appropriate Width Step. However, when I click on the top and drag the window (to move it in space), it seems to snap back to the previous size. I haven't noticed this with any of the other window-size-permutations.

 
- [as far as I can tell] This is a new bug that hasn't been documented in a closed issue or in the Readme.

macOS version: 26.0.1
Rectangle version: 0.96 (102)
Logs if applicable (In Rectangle menu, hold option, "View Logging..."):  

2026-06-29T14:31:54+03:00: AX sizing proposed: (1384.0, 831.0), result: (1384.0, 831.0)
2026-06-29T14:31:54+03:00: AX position proposed: (164.0, 73.0), result: (164.0, 73.0)
2026-06-29T14:31:54+03:00: AX sizing proposed: (1384.0, 831.0), result: (1384.0, 831.0)
2026-06-29T14:31:54+03:00: smaller, display: (0.0, 0.0, 1680.0, 1019.0), calculatedRect: (164.0, 73.0, 1384.0, 831.0), resultRect: (164.0, 73.0, 1384.0, 831.0), srcScreen: Built-in Retina Display, destScreen: Built-in Retina Display, resultScreen: Built-in Retina Display
2026-06-29T14:31:54+03:00: AX sizing proposed: (1354.0, 801.0), result: (1354.0, 801.0)
2026-06-29T14:31:54+03:00: AX position proposed: (179.0, 88.0), result: (179.0, 88.0)
2026-06-29T14:31:54+03:00: AX sizing proposed: (1354.0, 801.0), result: (1354.0, 801.0)
2026-06-29T14:31:54+03:00: smaller, display: (0.0, 0.0, 1680.0, 1019.0), calculatedRect: (179.0, 88.0, 1354.0, 801.0), resultRect: (179.0, 88.0, 1354.0, 801.0), srcScreen: Built-in Retina Display, destScreen: Built-in Retina Display, resultScreen: Built-in Retina Display
2026-06-29T14:31:56+03:00: AX position proposed: (196.0, 83.0), result: (196.0, 83.0)
2026-06-29T14:31:56+03:00: AX sizing proposed: (1414.0, 861.0), result: (1414.0, 861.0)

[...]

PR #1855, opened 2026-09-13
fix: keep the restore rect when unsnapRestoreFromSizeChange is off

Yash121l:
Related to #1784.

I went looking at #1784 and saw you already added `unsnapRestoreFromSizeChange` in 0.98, so this is not an attempt to change the default or add the settings checkbox. I know you want to do that with the UI rework. It is a smaller thing I hit while reading that code.

When `unsnapRestoreFromSizeChange` is set to 2, `getRestoreRect` returns nil for the size-change case, and `unsnapRestore` treats a nil there the same as "Rectangle never placed this window" and runs `restoreRects[windowId] = initialWindowRect`. So the drag correctly skips the resize, but it overwrites the user's stored restore rect with the resized frame.

Concretely: size a window yourself, press Make Smaller, drag it. The window stays small, which is what the setting is for, but the restore rect is now the small frame, so Restore brings back the small window instead of the size you had set.

I split the suppression check out into `isRestoreSuppressedBySizeChange` and return early from `unsnapRestore` before the branch that records a restore rect. `getRestoreRect` goes back to answering only "did Rectangle put this window here". `unsnapRestore` is internal again so the test can drive it; it was internal before d0a893d.

I also added `unsnapRestoreFromSizeChange` to TerminalCommands.md, since it shipped undocumented and it is the answer people land on #1784 looking for. If you would rather document it yourself, say so and I will drop that part.

Two tests in `SnappingManagerUnsnapRestoreTests` cover it, using the rects from the log in #1784. Full suite: 366 tests, 0 failures.

### DIFF
--- a/Rectangle/Snapping/SnappingManager.swift
+++ b/Rectangle/Snapping/SnappingManager.swift
@@ -317,8 +317,9 @@ class SnappingManager {
         }
     }
     
-    private func unsnapRestore(windowId: CGWindowID, currentRect: CGRect, cursorLoc: CGPoint?) {
+    func unsnapRestore(windowId: CGWindowID, currentRect: CGRect, cursorLoc: CGPoint?) {
         guard !Defaults.unsnapRestore.userDisabled else { return }
+        guard !isRestoreSuppressedBySizeChange(windowId: windowId) else { return }
         
         // if window was put there by rectangle, restore size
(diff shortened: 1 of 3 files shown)


rxhanson:
Nice, thanks! Gotta get my UI redesign out the door - aiming for this week.

PR #1856, opened 2026-09-13
Add an option to keep constrained windows top-aligned

### LINKED ISSUE #1825: Allow clamped side-snapped windows to retain top alignment

## Description

Since Rectangle v0.97, aspect-ratio-constrained windows are vertically centered when snapped to the left or right.

IINA is an easy example because its window preserves the video’s aspect ratio. **Left Half** correctly resizes and anchors it to the left edge, but now places it at the left-middle. Before v0.97, it remained at the top-left. **Right Half** similarly placed it at the top-right.

This also affects repeated half commands, which cycle through:

```text
1/2 → 2/3 → 1/3
```

Each size previously remained aligned with the corresponding top corner.

## Steps to reproduce

1. Open a widescreen video in IINA.
2. Run **Left Half**.
3. Observe that the window is vertically centered instead of top-aligned.
4. Repeat the action to cycle through the available widths.

## Expected behavior

Please provide an option to retain the snap area’s leading-corner position when a window cannot fill its requested height:

- Left Half → top-left
- Right Half → top-right

The current centered behavior can remain the default.

## Context and possible fix

This appears to come from the clamped-window alignment introduced in #1778 and enabled by default in #1789.

The current `moveFixedSizeToEdge` modes cover edge, corner-only, and centered alignment, but none retains the previous leading-corner position.

One possible fix would be a fourth mode that bypasses post-resize alignment and retains the position produced by `StandardWindowMover`:

```text
0 = retain leading-corner position
1 = align edges and corners
2 = align corners only
3 = center within the snap area
```

Alternatively, alignment on an unfilled axis could be made configurable as leading, center, or trailing.

If the legacy-mode approach is acceptable, I’d be happy to open a focused PR with geometry tests and a `TerminalCommands.md` update.

## Environment

```text
macOS: 26.6.2
Rectangle: 0.99 (105)
IINA: 1.4.4 (168)
```

Related: #1778, #1789, #809

slava-class:
## Summary

Add an option to keep constrained windows at the top of their snap area.

## Why another mode?

| Mode | Placement within the snap area |
| --- | --- |
| **1 — default** | Align to screen edges and corners; center where no single edge applies. |
| **2** | Align corner snaps; center other snaps. |
| **3** | Always center. |
| **4 — new** | Use the snap area's top-left origin. |

For left/right snaps, existing modes center shorter video windows vertically. Mode 4 keeps them top-aligned as you cycle through sizes.

- Applies to resizable and fixed-size windows; existing modes, the default, and screen containment are unchanged.
- 28 focused tests passed; Debug/arm64 build succeeded.

Closes #1825.

### DIFF
--- a/Rectangle/WindowMover/FixedSizeWindowMover.swift
+++ b/Rectangle/WindowMover/FixedSizeWindowMover.swift
@@ -10,14 +10,13 @@ class FixedSizeWindowMover: WindowMover {
         let currentWindowRect: CGRect = windowElement.frame
         if currentWindowRect.isNull { return }
 
-        let sharedEdges = Defaults.moveFixedSizeToEdge.value.alignmentEdges(
-            for: resultParameters.calcResult.initialRect.screenFlipped,
-            in: resultParameters.visibleFrameOfScreen.screenFlipped
+        let adjusted = ClampedWindowAligner.aligned(
+            window: currentWindowRect,
+            inZone: rect.screenFlipped,
+            initialRect: resultParameters.calcResult.initialRect.screenFlipped,
+            screenFrame: resultParameters.visibleFrameOfScreen.screenFlipped,
+            alignment: Defaults.moveFixedSizeToEdge.value
         )
-
-        let adjusted = ClampedWindowAligner.aligned(window: currentWindowRect,
-                                                    inZone: rect.screenFlipped,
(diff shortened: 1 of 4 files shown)


rxhanson:
Thanks for contributing this! The code looks good - I'll merge it in once I test it out manually in the next day or so.

PR #1857, opened 2026-09-14
tiling: Cycle Rows and Columns through balanced grids

dank-openai:
## Summary

- Repeated Rows and Columns actions cycle through balanced grid layouts using the same shortcuts.
- Separate limits for windows per row and windows per column appear beside the shortcut fields in General > Extras. Both default to 3; setting a limit to 1 disables cycling for that action.
- Tiling follows the existing display, Space, Todo, screen-edge gap, and optional-menu rules.

## Why This Change

Cycling lets users try different row or column arrangements with repeated presses of one shortcut. Each grid fills the available work area when application size restrictions allow it. Uneven window counts are balanced across occupied groups, and duplicate layouts are skipped.

For six windows with a maximum of three per column, successive layouts are:

1. Six columns of one window.
2. Three columns of two windows.
3. Two columns of three windows.
4. Six columns of one window again.

Rows cycle through six rows of one window, three rows of two, then two rows of three. Application size restrictions take precedence over equal sizing; infeasible layouts can still leave gaps or overlaps.

Moving, resizing, opening, or closing windows does not trigger automatic reflow. The next invocation reapplies the active layout to the current windows and positions. Repeating it without further changes advances. If applying a layout leaves all window frames unchanged, the same invocation tries the next layout. Cycle state lives only in memory; after a restart, an intact recognizable grid can still advance.

Explicitly saved limits remain intact, and the default of 3 applies only when a limit is unset. The two limits are independent of the single-window Repeated commands setting. No settings migration is required.

## Tests

37 focused XCTest cases passed, covering grid sequences and wraparound in both directions, uneven counts, window changes, pixel allocation, size restrictions, saved-limit round trips, window selection, and the optional Tiling menu. Window movement uses test doubles, and preference tests use isolated domains.


rxhanson:
Nice! The code looks good; I'll test this out soon.

PR #1858, opened 2026-09-14
Tweaked app icon for macOS 27

alexkaessner:
Tweaked the specular highlight for the new macOS 27 icon rendering. See the sharper 3 dots:
| Before | After |
|--------|--------|
| Before | After | 

PSA: From what I know this requires Xcode 27 to be used for the building.


rxhanson:
Nice, thanks, Alexander! I'll see if I can get the GitHub actions to run a successful build with macOS 27 before I merge so that it doesn't appear that the build is broken.

PR #1859, opened 2026-09-15
[Maintenance] Dev/blur preview

kiteretsu903:
Split a branch for dev of blur preview


rxhanson:
Alright, so after merging the other PR, I ended up walking the path of building on macOS 27 finally, and I relented on letting Xcode specify the minimum macOS version. That should be most of the merge conflicts, aside from me changing around a few things on the `WindowManager` that I'm certain caused a few. Sorry! Let me know if you want me to lend a hand on any parts of this.

kiteretsu903:
@rxhanson  Yep, resolved the conflicts

rxhanson:
Thanks! After spending a while with this and walking through the code and trying my own tweaks on what's here, I'm feeling inclined not to pull it in. I feel bad because I led you down this path, but I think the experience of your direct resizing is simply better even with the limitations it faces. Sometimes it's hard to figure things like this out without giving it a try, though. I appreciate your responsiveness and effort here! What are your thoughts?

kiteretsu903:
@ryonakano  Yep, i agree to drop this for now. it was a great experiment haha but it does look not ideal to use the window parking approach

rxhanson:
Sounds good. Again, I appreciate the work! It's always possible that this could be useful in some form in the future.

PR #1862, opened 2026-09-18
Preserve Welcome choices across Accessibility reauthorization

dank-openai:
## Summary
- Show Welcome only while first-launch onboarding is pending, preserving returning users' shortcut and repeated-action choices after Accessibility reauthorization.
- Keep Welcome pending across an exit before the first permission grant, including when permission is already granted at restart.

## Problem solved
The new-permission callback opens Welcome unconditionally. For example, a returning user who chose Resize can reauthorize Rectangle, see Welcome again, and select Recommended, changing the saved choice to Across Monitor.

Save that Welcome is pending before version bookkeeping records the first launch, so exiting before permission is granted does not suppress Welcome on restart. Both newly granted permission and an already-trusted launch check that pending state. Clear the marker after the Welcome callback returns. Existing installations retain their preferences without a migration.

## Validation
- Added three focused regression tests: returning-user Resize preservation, new-user Welcome, and interrupted first launch. The latter two exercise both authorization paths.
- All three pass. Restoring the old authorization handling reopens Welcome for returning users and skips it for new or interrupted first launches when permission is already granted. Preventing creation of the pending marker also makes the new-user and interrupted-launch tests fail.
- Unsigned Debug application and focused test builds pass. The focused test bundle was built with `EXCLUDED_SOURCE_FILE_NAMES=RectangleTests.swift`: that existing file has an unrelated outdated `setFrame` override that prevents the full suite from compiling. No exclusion or unrelated test repair is included in this PR.
- Tests use isolated UserDefaults suites and fake authorization and Welcome callbacks. The returning-user test models selecting Recommended without opening UI. Restart is simulated by recreating the controller and preferences object; real process restart, Accessibility permission, and modal UI were not exercised.


dank-openai:
This issue cropped up while I was developing, so I found it useful to fix it.

rxhanson:
Thanks for the follow-on. 

To keep it simpler, I think we could instead add this to the beginning of the `showWelcomeWIndow` function:

```
        if Defaults.hasShownWelcomeWindow.enabled {
            return
        }
        Defaults.hasShownWelcomeWindow.enabled = true
```

And add that BoolDefault to Defaults.

```
    static let hasShownWelcomeWindow = BoolDefault(key: "hasShownWelcomeWindow")
```

There isn't a need to make that part unit testable, in my opinion, so we could get by with this simple change.

dank-openai:
> There isn't a need to make that part unit testable, in my opinion

Simplified this into AppDelegate and a BoolDefault, removing the dedicated controller and tests. I kept the pending-Welcome flag so existing users without the new flag do not see Welcome again on reauthorization. The flag is saved before the app records its first launch, so quitting before granting permission does not lose the pending Welcome. It is checked both when permission is granted and when the app starts with permission already granted.

rxhanson:
Not desiring to split hairs here, but the `Defaults.lastVersion.value == nil` check is not necessary, right? We can simply check if the welcome window has been displayed when we try to display it, as mentioned in my previous comment. (I'd like to clean up the AppDelegate at some point so I'm interested in not adding unnecessary logic to it)

rxhanson:
I'm in the midst of rewriting a lot of parts of the app to use SwiftUI, and decided to go ahead and rewrite the Welcome window at this time. As part of that, I'll go ahead and absorb the check for whether or not it's been displayed into the welcome window itself. Thanks for influencing the direction of the app to accommodate this!

